PR #1, opened 2019-05-04
fix upstream actions being sent to SideEffects twice

gabrielittner:
The issue was that we were sending the action to the BroadcastChannel twice. Once in the removed `broadcastIn` and another time inside the upstream `collect` call. I decided to remove `broadcastIn` because this way the upstream and sideEffect collect calls are identical and it makes it clearer that the reducer is called before we are sending the action to the side effects.

### DIFF
--- a/library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt
+++ b/library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt
@@ -1,21 +1,15 @@
 package com.freeletics.flowredux
 
 import kotlinx.coroutines.channels.BroadcastChannel
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.asFlow
-import kotlinx.coroutines.flow.broadcastIn
 import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.flow.delayEach
-import kotlinx.coroutines.flow.delayFlow
 import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.flatMap
 import kotlinx.coroutines.flow.flow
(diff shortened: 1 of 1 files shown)


gabrielittner:
Output of main:
```
Emitting initial state
STATE: Start
Subscribing sideeffect
Received Action 1 from upstream
STATE: Start1
-- SF1: Got action 1 . current state Start1
Got action 3 from sideeffect
STATE: Start13
-- SF1: Got action 3 . current state Start13
Received Action 2 from upstream
STATE: Start132
-- SF1: Got action 2 . current state Start132
Got action 3 from sideeffect
STATE: Start1323
-- SF1: Got action 3 . current state Start1323
```

gabrielittner:
This introduces a different issue. main will never return because the `launch` inside `reduxStore` never completes. Just doing it the other way around (keeping `broadcastIn` and removing the send inside the upstream `collect`) still has the issue of subscribing twice to the upstream which creates two separate Flows.

For example when changing the flow in main to use a shared resource
```
val counter = AtomicInteger()
flow {
    delay(1000)
    emit(counter.incrementAndGet())
    delay(4000)
    emit(counter.getAndIncrement())
}
```
and deleting this line to fix receiving each action twice in side effects:
https://github.com/freeletics/FlowRedux/blob/a9209254113ded2a297ecd3d1e294836bdcd13e3/library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt#L74

You'd get this:
```
Emitting initial state
STATE: Start
Subscribing sideeffect
Received Action 1 from upstream
STATE: Start1
-- SF1: Got action 2 . current state Start1
Got action 3 from sideeffect
STATE: Start13
-- SF1: Got action 3 . current state Start13
Received Action 2 from upstream
STATE: Start132
-- SF1: Got action 3 . current state Start132
```


I'll create a new PR which will:
1) fix receiving actions twice in side effects (broken on master)
2) only subscribe to the upstream once (broken on master)
3) still completes when upstream completes (works on master)

PR #5, opened 2019-05-06
 only subscribe once to upstream, properly terminate

gabrielittner:
This fixes:
- receiving actions twice in side effects
- subscribing to the upstream twice
- returned Flow does not always complete when upstream completes

Open issues:
- SideEffects are called in a weird order (see the last two tests which are failing).
- When upstream completes SideEffects are not run on actions emitted by SideEffects in response to the last upstream action. So if upstream emits a `2`, the `2` is going through the reducer and the SideEffects. When a SideEffect emits an action in response to the `2` that action will go through the reducer, but not again to SideEffects. The cause for this is the BroadcastChannel being closed so send isn't working anymore. I'd say that this isn't a big issue because for redux use cases the upstream usually never completes.

### DIFF
--- a/library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt
+++ b/library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt
@@ -86,8 +86,8 @@ fun <A, S> Flow<A>.reduxStore(
 
 @UseExperimental(FlowPreview::class)
 fun main() = runBlocking {
-    val sideEffect1: SideEffect<String, Int> = { action: Flow<Int>, stateAccessor: StateAccessor<String> ->
-        action.flatMapConcat { action ->
+    val sideEffect1: SideEffect<String, Int> = { actions, stateAccessor ->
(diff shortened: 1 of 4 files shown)


library/src/main/kotlin/com/freeletics/flowredux/FlowRedux.kt
```diff
@@ -42,18 +35,19 @@ fun <A, S> Flow<A>.reduxStore(
 
     coroutineScope {
 
-        val actionBroadcastChannel: BroadcastChannel<A> = this@reduxStore.broadcastIn(this)
+        val actionBroadcastChannel: BroadcastChannel<A> = BroadcastChannel(100)
```
gabrielittner: Not sure about what `capacity` here is actually appropriate. `broadcastIn` uses 1, but that's not enough for us.

sockeqwe: But do we need to broadcast more than 1 action at a time?

PR #126, opened 2021-06-16
implement Freeletics MAD StateMachine

gabrielittner:


.github/workflows/publish-release.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.MAVEN_SIGNING_PASSWORD }}
 
       - name: Upload release (Windows)
-        run: ./gradlew publishMingwPublicationToMavenRepository --no-daemon --no-parallel
+        run: ./gradlew publishMingwX64PublicationToMavenCentralRepository --no-daemon --no-parallel
```
gabrielittner, line 42: Was a mistake I made in the initially added file in the mad repo. That's why there is a `0.1.1` release which has the missing windows artifacts where publishing failed. The snapshot workflow was already correct.

build.gradle
```diff
@@ -23,8 +23,6 @@ allprojects {
         mavenCentral()
         maven { url "https://jitpack.io" }
         maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
-        maven { url "https://kotlin.bintray.com/kotlinx" }
-        maven { url "https://dl.bintray.com/jetbrains/kotlin-native-dependencies" }
```
gabrielittner, line 27: Do not exist anymore because of the bintray shutdown

dsl/build.gradle
```diff
@@ -83,6 +84,7 @@ kotlin {
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     kotlinOptions {
         freeCompilerArgs += ["-Xuse-experimental=kotlin.Experimental"]
+        jvmTarget = "1.8"
```
gabrielittner, line 87: Needed because flMadStateMachine is built with `1.8`.

sample/android/build.gradle
```diff
@@ -90,7 +90,7 @@ dependencies {
     implementation libraries.okhttp
     implementation libraries.moshiKotlin
     implementation libraries.moshi
-    implementation("io.ktor:ktor-client-okhttp:1.3.2")
+    implementation("io.ktor:ktor-client-okhttp:1.3.1")
```
gabrielittner, line 93: Dependency couldn't be found with 1.3.2. I aligned it with the other ktor versions before which work

sample/android/build.gradle
```diff
@@ -120,7 +120,7 @@ dependencies {
     androidTestImplementation testLibraries.espresso
     androidTestImplementation testLibraries.espressoContrib
     androidTestImplementation testLibraries.testRules
-    androidTestImplementation testLibraries.screengrab
+    // androidTestImplementation testLibraries.screengrab
```
gabrielittner, line 123: Does not exist anymore because of the bintray shutdown, just commented it out for now so that the library can be built and the sample build was failing before already.

build.gradle
sockeqwe: 👍  Thanks
Overall this repository needs a lot of clean up, most of the things were hacked together and sample was copied from RxRedux but evolved overtime so there are still some unused dependencies.

So thanks for taking the first step 👍

dsl/build.gradle
sockeqwe: 👍

sample/android/build.gradle
sockeqwe: Yea, this is one of the examples of not used dependencies that were copy and pasted over when forking from RxRedux.

The sample needs to be cleaned up.

sockeqwe (approved):

PR #150, opened 2021-07-04
make FlowReduxStateMachine based on StateFlow

### LINKED ISSUE #130: StateFlow and CoroutineScope as constructor Parameter
For `FlowReduxStateMachine` we should use `StateFlow` instead of Flow so that the StateMachine can work independent from it's state subscribers (like UI can subscribe and unsubscribe according to UI's lifecycle. 

Furthemore, the statemachine should keep alive independent from subscribers lifecycle. Therefore a CoroutineScope should be passed as constructor parameter. That CoroutineScope will then be used to keep the whole statemachine alive until that scope gets canceled. On canceling this statemachine firther calls of `dispatch()` or `state.collect()` must throw an `IllegalStateException`.

gabrielittner:
Closes #130

### DIFF
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
@@ -1,34 +1,29 @@
 package com.freeletics.flowredux.dsl
 
 import com.freeletics.flowredux.FlowReduxLogger
+import com.freeletics.mad.statemachine.StateMachine
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.onSubscription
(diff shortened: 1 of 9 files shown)


dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -1,34 +1,29 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.onSubscription
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
 
 @ExperimentalCoroutinesApi
 abstract class FlowReduxStateMachine<S : Any, A : Any>(
-    logger: FlowReduxLogger?,
-    initialStateSupplier: () -> S
```
gabrielittner, line 12: I removed the supplier because we'd call it immediately anyways.

dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StateMachine.kt
```diff
@@ -1,16 +1,20 @@
...
 fun StateMachine(
         builderBlock: FlowReduxStoreBuilder<TestState, TestAction>.() -> Unit
 ): FlowReduxStateMachine<TestState, TestAction> {
-    val sm: FlowReduxStateMachine<TestState, TestAction> = object : FlowReduxStateMachine<TestState, TestAction>(
-            CommandLineLogger,
-            TestState.Initial){
-
+    val sm = object : FlowReduxStateMachine<TestState, TestAction>(
+        TestState.Initial,
+        CoroutineScope(Dispatchers.Unconfined),
```
gabrielittner, line 15: When I used `Dispatchers.Default` here some tests fail. I'm not sure if this is just because of how the tests are written or hints at some problem in the implementation of the DSL code.

Issue #128, opened 2021-06-18
Cancel on<Action> job when surrounding inState condition doesn't anymore


sockeqwe:
At the moment we don't cancel any ongoing handler of a `on<Action>` if the surrounding `inState<State>` condition doesn't hold anymore (so statemachine transitioned to another state).

This was actually the expected behavior before FlowRedux 0.5.0, but now should change.

gabrielittner:
Doesn't this test mean that we do cancel? https://github.com/freeletics/FlowRedux/blob/main/dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/OnActionTest.kt#L48 If we wouldn't cancel during the delay then `reached` would be `true`. Is there some other case where we don't cancel?

PR #163, opened 2021-07-22
add support for inState effects

### LINKED ISSUE #129: Add support for Effects that dont change state
At the moment you cannot do a piece of work without the need of returning a `ChangeState` object.

This proposal is about adding "effects". A typical use case for an  Effect is triggering navigation without changing the state of the statemachine itself or doing some analytics tracking calls.

Proposed Syntax
```kotlin
inState<MyState> {
  onEnterEffect { stateSnapshot : MyState ->
       // Do something.
       Unit // return unit, no state change
   }

  onActionEffect<MyAction> action : MyAction, stateSnapshot : MyState ->
       // Do something.
       Unit // return unit, no state change
   }

  collectWhileInStateEffect(flowOf(1,2)) { value : Int, 
stateSnapshot : MyState ->
       // Do something.
       Unit // return unit, no state change
   }
}
```

Specs:
- Effects are suspending: i.e. signature lambda for `onEnterEffect ` is `suspend (State) -> Unit`
- Effects dont get canceled when state transition happening.
- Effects are per default canceled when StateMachine  CoroutineScope gets canceled. If you want to escape you can always use `withContext(NonCancelable)`

gabrielittner:
part of #129

From the issue:
> Effects dont get canceled when state transition happening.

The `collectWhileInStateEffect` will get cancelled after state changes otherwise the name wouldn't fit. With the current implementation the other 2 would not get cancelled, but they would after #128 is fixed, so we need to consider that when working on that bug.


sockeqwe:
I would prefer consistent behavior over consistent naming ... So renaming collectWhileInStateEffect to something else would be fine for me

gabrielittner:
But wouldn't it make sense if gets cancelled? If we don't cancel collect it would be a "start collecting when entering this state and never stop" which to me seems like a footgun. The only way to cancel it would be from within the stream.

sockeqwe:
Yea, I think we want to have a way to do some work without changing the
state. I think, however, there are different use cases and we need to be
clear which we want to cover here.

1. Trigger something like navigation. I think this is our main use case and
what you had in mind with this PR. In that case maybe we actually want to
cancel such a side effect if state changed.
2. I was originally thinking a bit mor of a long living effect that could
outlive state transitions but also maybe even run on another
CoroutineScope. Usecase: write current state to disk storage or do an http
request that doesn't cancel.

I thought the second one would be a generalization of the first one in the
sense covering our navigation use case ...

What do you think?

Gabriel Ittner ***@***.***> schrieb am Fr., 23. Juli 2021,
10:38:

gabrielittner:
When we talk about not cancelling `collectWhileInStateEffect` are we talking about not cancelling an already executing lambda or about not cancelling the collection itself. I was talking about cancelling the collection itself. If we don't cancel it the lambda wouldn't work with the substate snapshot because the state can be different.

sockeqwe:
oh, you are right, maybe we were talking about different things.

I am talking about cancelling the underlying collection of the flow, but not cancelling the lambda that is executed.

example:

```kotlin
val sampleFlow = flow {
     emit (1)
     delay(1000)
     emit(2)
}

... 


spec {
    inState<FooState> {
       collectWhileInStateEffect(sampleFlow) { item, state ->
           delay(2000)
           saveToDisk(state)
       }
    }
    
    inState<OtherState> { ... }
}
```

then let's say the following things are happening:
1. StateMachine enters `FooState` 
2. `collectWhileInStateEffect` subscribes to `smapleFlow`
3. Lambda of `collectWhileInStateEffect` is executed with  `item = 1, state = FooState`. This lambda executes and runs into the `delay(2000)`.
4. After 500 ms, StateMachine moves into `OtherState`. So now, collecting the `sampleFlow` should stop. so `sampleFlow` would not `emit(2)`
5.  The lambda which execution has started in Step 3, however, should NOT be canceled, so saveToDisk(state) should still continue to execute even if the statemachine has moved to `OtherState` in the meantime


Does that make sense for you?

What would be your expected behavior?

gabrielittner:
What you describe is what I would expect to happen 👍

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/MapStateChange.kt
```diff
@@ -1,100 +0,0 @@
...
-import kotlinx.coroutines.flow.filter
-import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
-
-/**
- * Internal implementation of an operator that keeps track if you a certain state has been
- * entered or left.
- */
-internal class MapStateChange<S : Any, A : Any>(
```
gabrielittner, line 15: Was unused after the previous refactorings

gabrielittner:
@sockeqwe Updated the PR with new implementations of `onActionEffect` and `onEnterEffect` that have the intended cancellation behavior. 

I will tackle `collectWhileInStateEffect` separately but it's are going to be a lot more complex. The main issue there is that we want to cancel the upstream flow but do not want to cancel a running handler. I'm not even entirely sure if completely cancelling is even possible I think that will always cancel everything. `takeWhile` would work but the upstream would stay active until emits it's next value, from the state machine behavior everything would then work as intended but the upstream flow would stay active longer than necessary.

sockeqwe:
Since it is unclear how it could be implemented consistently we can discuss to go back to just work as `onEnter`, `onAction` and `collectWhileInState` do and make the "Effects" counterpart basically just syntactic sugar and return `NoStateChange`.

I would propose to go with this strategy for now and revisit the cancellation start if we then see enough use case for Effects to  outlive `inState` scope.

What do you think?

gabrielittner:
Yes, makes sense

PR #167, opened 2021-07-29
remove collectWhileInAnyState

### LINKED ISSUE #162: Inconsistency between collectWhileIn(Any)State and on
As an example let's say we have this state
```
sealed class MyState
class StateA : MyState()
class StateB : MyState()
```

Now in a state machine for `MyState` you can do
```
init {
    spec {
        collectWhileInAnyState(...) { ... }
        
        inState<MyState> {
            collectWhileInState(...) { ... }
            on<...>() { ... }
        }
    }
}
```

The 2 `collectWhile...` calls have the same lifetime because we are always in `MyState`. The inconsistency in my opinion is that there are effectively 2 ways to do `collectWhileInAnyState` but it's only possible to define an `on` that should happen in any state within `inState`. Initially I was searching for an `onInAnyState` because I knew `collectWhileInAnyState` exists. 

Instead of having `collectWhileInAnyState` could we just remove it and recommend using `collectWhileInState` with the base class? Then it's consistent with `on` and there is just one way to do things.

gabrielittner:
closes #162 

I've also done the change to move the internal parts of the dsl module into a subpackage for easier readability that we talked about.


docs/dsl.md
```diff
@@ -485,53 +485,49 @@ otherwise not (returning false). The rest still remains the same. You can use `o
...
+    // Shows a loading indicator on screen
+    object LoadingState : State()
+
+    sealed class PostLoadingState : State()
+
+    // List of items loaded successfully, show it on screen
+    data class ShowContentState(val items: List<Item>) : PostLoadingState()
+
+    // Error while loading happened
+    data class ErrorState(val cause: Throwable) : PostLoadingState()
```
sockeqwe, line 527: Nice idea 👍

sockeqwe (approved):
LGTM 👍

PR #168, opened 2021-07-29
fix tests to some degree

gabrielittner:
The tests were generally not running correctly because runBlocking is not transitive. So if we `launch` inside the test to call `sm.state.test {}` nothing waits for that test block to finish. It just gets cancelled. This is most noticeable in the places where we had `expectComplete()`. The `Flow` returned by `reduxStore` (even without DSL) never completes, which shows those tests didn't actually fully run before.

The bad news is that the cancellation tests still work even though they shouldn't. Switching the dispatcher does break them but it also breaks tests that shouldn't break.I will look into that more.


sockeqwe (approved):
👍

PR #169, opened 2021-07-29
fix issues hidden by Unconfined Dispatcher

### LINKED ISSUE #128: Cancel on<Action> job when surrounding inState condition doesn't anymore

### LINKED ISSUE #171: onEnter doesn't need FlatMapPolicy
Since `onEnter(FlatMapPolicy)` is exactly triggered once (and doesnt emit later on again) it doesnt need a FlatMapPolicy as a parameter

gabrielittner:
This replaces the dispatcher used in tests with `Dispatchers.Default`. There was a combination of things at play here.

1. Initial state is now visible to all tests again, this was expected from me. I only removed those expects in #130 because I had to use Unconfined
1. Any test that relied on a dispatched action for the first state change (`OnActionTest` and `CustomIsInStateDslTest`) was affected by an issue in `FlowReduxStateMachine`. Because of the launch that is done to start the store and it using `MutableSharedFlow` for incoming actions, any action that is dispatched before the launch body executes was lost. So the tests relying on an action didn't work. In real world use cases this is probably not an issue because you don't start dispatching actions a few millisecods later. I've changed the implementation to a `Channel` which will suspend the `dispatch` until we start collecting the actions.
1. The updated onEnter test was not set up for asynchronous events. `expectNoEvents` does not suspend or anything to wait that there are no more events, it just checks that there is nothing in it's queue. Because of the async dispatch of `A1` the following `assertEquals` happens too early and the counter is still 0 because not even `onEnter` of `S1` finished running yet. I've changed the test to use `GenericState` witch changing values so that we can wait for them with `expectItem` while still staying in the same state. The assertion that `onEnter` was only called once now happens after the `test` block when we know all invocations are done.

fixes #128, closes #171

### DIFF
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -576,10 +576,9 @@ could complete before the first execution (because using a random time of waitin
 - `CONCAT`: In contrast to `MERGE` and `LATEST` `CONCAT` will not run `on<BarAction>` in parallel and will not cancel
   any previous execution. Instead, `CONCAT` will preserve the order and execute one block after another.
 
-All execution blocks can specify a `FlatMapPolicy`:
+All execution blocks except `onEnter` can specify a `FlatMapPolicy`:
 
 - `on<Action>(flatMapPolicy = FlatMapPolicy.LATEST){... }`
-- `onEnter(flatMapPolicy = FlatMapPolicy.LATEST) { ... }`
 - `collectWhileInState(flatMapPolicy = FlatMapPolicy.LATEST) { ... }`
 
 ## Best Practice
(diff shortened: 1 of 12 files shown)


sockeqwe (approved):
Looks good, still not sure why tests are failing

gabrielittner:
Everything's greed 🎉

Issue #166, opened 2021-07-29
collectWhileInState doesnt cancel when leaving the state


sockeqwe:

sockeqwe:
Fixed with #170

PR #170, opened 2021-07-29
break OnAction and OnEnter canellation tests

gabrielittner:
Builds on top of #169.

With all the other stuff out of the way I was now finally able to reproduce #128 in tests. Now that it's running asynchronously all that needed to be done is moving the `assertFalse` for reached down. Similar to case 3 in #169 the assertion happened before the action was processed in the state machine so `reached` was still `false` but it would change a bit later. It would have been enough to move them one line down after the next `expectItem` but I moved them to the end where they should always work.

Unfortunately I still can't break a test for #166/collectWhileInState.


gabrielittner:
> Unfortunately I still can't break a test for #166.

I said that, but in the other pr (#169) the test collectWhileInState test is failing :D

sockeqwe (approved):

sockeqwe:
Lets merge all these back into `test-unconfined` branch and use that one as base branch to fix things

PR #172, opened 2021-07-31
fix race condition in reduxStore

gabrielittner:
Builds on top of #169. The issue is basically the same as point 2 described in that pr, but this time the race condition is directly in the reduxStore operator.

The implementation of the dsl will emit the internal InitialStateAction in onStart. The issue was that this in some cases didn’t reach the side effects after going through the reducer. The reason for that is a race condition when setting up all the flows. The merge operator will internally do a launch for each given Flow and start collecting them in there. That works fine and is ordered. However I was able to observe that the collect going up the side effect chain can sometimes be slower than the subscription to the upstream actions.

In that case the timeline was:
1. start collecting sideeffects and upstream
2. reducer receives InitialStateAction and emits it to loopback
3. side effects start collecting loopback

Since loopback is just a broadcast to all current subscribers, InitialStateAction just gets lost in this case, which breaks any onEnter or collectWhileInState for the initial state. In even trickier cases it could also happen that 3 partially happens before 2 so some side effects get InitialStateAction while others don’t.

I think it didn’t happen before starting to use Dispatchers.Default because the setup was basically single threaded and then 3 would always happen before 2 and everything is fine.
The solution is to give each side effect a channel to read from. With this implementation the it.send(action) will suspend until the corresponding side effects starts collecting it. So if the race condition happens the action will not be dropped anymore.

I can't reproduce the ci failure of `com.freeletics.flowredux.dsl.CollectWhileInStateTest[jvm] > collectWhileInState stops after having moved to next state[jvm] FAILED` locally (ran each test 1k times in a loop), but it already failed in the unconfined pr so maybe it's the cancellation not working bug that you saw before.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/FlowRedux.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/FlowRedux.kt
@@ -1,9 +1,10 @@
 package com.freeletics.flowredux
 
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.onEach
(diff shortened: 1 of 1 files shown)


sockeqwe (approved):
Thanks for looking into it. 

After my (very short and not really in depth) investigation on Friday, I also thought that some race conditions could happen here.

Thanks for investigating 👍

sockeqwe:
I was in last spikeweek thinking of actually replacing the internals of this `.reduxStore()` operator with something only channel backed and get rid of "side effects concept ported from RxRedux". I was thinking about something like that (pseudocode, won't work obviously, just to sketch out my thoughts):


```kotlin
internal sealed interface Notification

internal data class StateChangeNotification(val changeState : ChangeState<T>) : Notification
internal data class IncomingActionNotification(val action : A) : Notification
... 

// DSLBlock is some sort of common interface for everything that you can build with DSL,
//  i.e. CollectWhileInStateDslBlock will be returned from collectWhileInState { ... }
internal interface DslBlock {
    fun isInState(state : S) -> Boolean 
    var job : Job?
    fun handle(notificationChannel : Channel<Notification>, stateSnapshot : S,  action : A? )  : Job
}

... 
fun Flow<S>.reduxStore {
var stateFlow = MutableSharedFlow()
val allBlocksFromDSL : List<DslBlock> = runDslBlocksToInstantiateBlocks(notificationChannel) 
val notificationChannel = Channel<Notification>()

for (notification in NotificationChannel) {
  when (notification) {
      is IncomingActionNotification ->  allBlocksFromDSL.forEach {
               it.job?.cancel()
               it.job = launch  {
                   it.handle(notificationChannel, stateFlow.state, notification.action)
               }
           } 

       is StateChangeNotification -> allBlocksFromDSL.forEach { 
              val oldState =  stateFlow.state
              val newState = it.changeState.reduce(oldState)
              stateFlow.state = newState
               ...
               // check which dslblocks are still in the state and either cancel ther jobs or if they now enter the state than startstart their jobs
            if (it.isInState(newState)){
                if (it.job == null) 
                   it.job = it.handle(notificationChannel, newState, null) 
                // else do nothing because if job already runs then the state didnt change for this DSL block so not restarting it
           ) else it.job?.cancel() // we exit the state that originally did start this DSLBlock, so cancel all ongoing action
        }
   }
}
```

 return stateFlow
}

PR #173, opened 2021-08-01
deduplicate CollectInState code, update name

gabrielittner:
Each when branch was identical except for the inner `flatMap...` call. Since we already have `flatMapWithPolicy` I've changed the inner flatMap to that and removed the when.

Also updates the name of the class because it's the only implementation by now.


sockeqwe (approved):
👍

PR #174, opened 2021-08-01
remove onEnter flatMapPolicy and fix cancellation when leaving state

### LINKED ISSUE #171: onEnter doesn't need FlatMapPolicy

gabrielittner:
Closes #171 and adresses the onEnter part of #128.

The old `flatMapWithPolicy` didn't get any events when leaving the state which is why it didn't cancel even if it was using `LATEST`. The new implementation turns each action emission to an `isInState` boolean and then has a `flatMapLatest` that will only return the onAction Flow if the boolean was true. This way the previous `Flow` will get cancelled when `isInState` changes from true to false. `distintUntilChanged` is used to ignore any additional emissions after the first `true`.

Instead of `map` to a boolean and `distintUntilChanged` this could also use `filterState` from the `CollectWhileInState` implementation. However this seems a lot less complex to me and I can't think of any issues with this implementation.

### DIFF
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -576,10 +576,9 @@ could complete before the first execution (because using a random time of waitin
 - `CONCAT`: In contrast to `MERGE` and `LATEST` `CONCAT` will not run `on<BarAction>` in parallel and will not cancel
   any previous execution. Instead, `CONCAT` will preserve the order and execute one block after another.
 
-All execution blocks can specify a `FlatMapPolicy`:
+All execution blocks except `onEnter` can specify a `FlatMapPolicy`:
 
 - `on<Action>(flatMapPolicy = FlatMapPolicy.LATEST){... }`
-- `onEnter(flatMapPolicy = FlatMapPolicy.LATEST) { ... }`
 - `collectWhileInState(flatMapPolicy = FlatMapPolicy.LATEST) { ... }`
 
 ## Best Practice
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
@@ -59,13 +59,11 @@ class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
      * TODO add a sample
      */
     fun onEnter(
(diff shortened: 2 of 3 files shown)


docs/dsl.md
```diff
@@ -576,10 +576,9 @@ could complete before the first execution (because using a random time of waitin
 - `CONCAT`: In contrast to `MERGE` and `LATEST` `CONCAT` will not run `on<BarAction>` in parallel and will not cancel
   any previous execution. Instead, `CONCAT` will preserve the order and execute one block after another.
 
-All execution blocks can specify a `FlatMapPolicy`:
+All execution blocks except `onEnter` can specify a `FlatMapPolicy`:
```
sockeqwe, line 579: Thanks for updating!

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -59,13 +59,11 @@ class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
      * TODO add a sample
      */
     fun onEnter(
-        flatMapPolicy: FlatMapPolicy = FlatMapPolicy.LATEST,
```
sockeqwe, line 62: 👍

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/OnEnterInStateSideEffectBuilder.kt
```diff
@@ -26,10 +26,14 @@ class OnEnterInStateSideEffectBuilder<InputState : S, S : Any, A : Any>(
     override fun generateSideEffect(): SideEffect<S, Action<S, A>> {
         return { actions: Flow<Action<S, A>>, getState: GetState<S> ->
             actions
-                .mapStateChanges(isInState = isInState, getState = getState)
-                .filter { it == MapStateChange.StateChanged.ENTERED }
-                .flatMapWithPolicy(flatMapPolicy) {
-                    setStateFlow(getState)
+                .map { isInState(getState()) }
```
sockeqwe, line 29: Thanks, this is a mich simpler implementation 🙏

sockeqwe (approved):

PR #175, opened 2021-08-01
cancel onAction when leaving the current state

### LINKED ISSUE #128: Cancel on<Action> job when surrounding inState condition doesn't anymore

gabrielittner:
Closes #128 

This one is more complex then fixing onEnter. Just the boolean that we have there because we still need the action value and new actions while we are in the state should not cancel the previous flow. We basically want a `flatMapSometimesLatest`. The new operator will put new upstream actions into a `Channel` when `isInState` is true. This `Channel` is then turned into a Flow and we apply our filtering and transformation with the flatMapPolicy on it. The resulting `Flow` gets collected and it's emissions are emitted to the downstream. When `isInState` becomes false the channel is closed so that the resulting flow will get cancelled as well.

### DIFF
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
@@ -1,0 +1,46 @@
+package com.freeletics.flowredux.dsl.flow
+
+import com.freeletics.flowredux.GetState
+import com.freeletics.flowredux.dsl.internal.Action
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.channelFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.launch
+
+@ExperimentalCoroutinesApi
+internal fun <S, A> Flow<Action<S, A>>.whileInState(
+    isInState: (S) -> Boolean,
+    getState: GetState<S>,
+    transform: (Flow<Action<S, A>>) -> Flow<Action<S, A>>
+) = channelFlow {
+    var currentChannel: Channel<Action<S, A>>? = null
+
+    // collect upstream
+    collect { value ->
+        if (isInState(getState())) {
+            // if there is no Channel yet the state machine just entered the state
+            if (currentChannel == null) {
+                currentChannel = Channel()
+                launch {
+                    // transform actions sent to the the channel with the given function
+                    // and emit the result downstream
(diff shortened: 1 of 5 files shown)


gabrielittner:
I broke something during the rebase, will look into it.

gabrielittner:
It's fixed now the test failures are the already failing onEnter and CustomIsInStateDslTest tests

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.launch
+
+@ExperimentalCoroutinesApi
+internal fun <S, A> Flow<Action<S, A>>.whileInState(
+    isInState: (S) -> Boolean,
+    getState: GetState<S>,
+    transform: (Flow<Action<S, A>>) -> Flow<Action<S, A>>
+) = channelFlow {
```
sockeqwe, line 21: Why do we need a ChannelFlow here?. wouldn't it work with just a regular Flow as well?

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
```diff
@@ -0,0 +1,46 @@
...
+    var currentChannel: Channel<Action<S, A>>? = null
```
sockeqwe, line 22: Do we need a mutex to ensure no 2 actions are processed in parallel or is there auch a guarantee already in place under the hood by channelFlow?

sockeqwe (approved):
Overall LGTM and your implementation makes sense to me. Left 2 questions, but no blocker ...

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
gabrielittner: The reason is that we emit/send from within the `launch`. Just using `flow { coroutineScope { ... } }` violates the context preservation requirement of flow and crashes, you can see that in the commit from before the force push

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/flow/WhileInState.kt
gabrielittner: Flow itself is sequential and under the assumption that collectors are not thread safe. So when we collect the upstream here we only get the next action when the previous one was handled which happens after we `send` it into `currentChannel`. So we sequentially put them into the channel and it also applies to creating/closing the channel. For the processing itself we are collecting the channel as a flow again, so it's als sequential and ordering of the action processing then depends on the transform. For onAction specifically it depends on the `flatMapPolicy`, CONCAT and LATEST are not parallel, while MERGE is.

PR #180, opened 2021-08-02
Update to AGP 7.0.0

sockeqwe:


gabrielittner (approved):

.github/workflows/build.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        java_version: [1.8, 11]
+        java_version: [11]
```
sockeqwe, line 12: We could remove Martix builds, but I like the possibility to add easily other JDKs, thus would keep it

PR #182, opened 2021-08-03
add collectWhileInState functions with flow builders

### LINKED ISSUE #161: collectWhileInState based on the current state
While my state machine is in a specific state I want to collect a query in my database. However the query has a filter that can change based on actions. I'd like to make that filter part of my state so that actions just change the state and then the query can update automatically. However for that I would need a `Flow` of the current state.

I'm currently working around this by having this in my state machine:
```
private val openingTime = MutableStateFlow(Instant.now())

init {
    spec {
        inState<MyState> {
            collectWhileInState(loadData(), block = ::updateStateWithData)
            on(block = ::onRefresh)
        }
    }
}

private fun loadData(): Flow<Feeds> {
    // here is where I'd like to have a `Flow<MyState>` instead
    return openingTime.flatMapLatest { repository.load(it) }
}

private fun updateStateWithData(...): ChangeState<MyState> { ... }

private suspend fun onRefresh(
    action: Refresh,
    state: MyState
): ChangeState<MyState> {
    // this should mutate MyState instead of updating an external Flow
    openingTime.emit(Instant.now())
    return NoStateChange
}
```
All of that feels very wrong with how the DSL based state machines work. Additionally if I need that `Instant` in my state for other purposes, like showing it in the UI, there would be 2 sources of truth for it and a chance of inconsistencies. For example if there'd be a second action that updates it I could forget to call `openingTime.emit` and then the UI would something different than what is being queried.

Maybe `collectWhileInState` should take a lambda that returns the `Flow` instead of directly receiving that `Flow` and we pass a `Flow` of the state to that lambda. So something like
```
fun <T> collectWhileInState(
    flow: (Flow<InputState>) -> Flow<T>,
    flatMapPolicy: FlatMapPolicy = FlatMapPolicy.CONCAT,
    block: InStateObserverBlock<T, InputState, S>
)
```
The lambda would also fit the style of using method references inside `spec` and maybe discourage a bit to build the `Flow` directly when calling `collectWhileInState` instead of in an extra function.

Note: With the change from #150 it would technically be possible to observe the main `state` of the `FlowReduxStateMachine` since it is visible to us and always collectable, but that feels also wrong. Especially in an `inState` block because `state` has all states and not just the filtered one.

### LINKED ISSUE #179: Add overloads for even nicer DSL with function reference
The problem is that parameter with default args doesnt play nice with function references as it requires named arguments.

i.e.
```kotlin 
inState<S1> {
   onAction<A1> (block = ::handlerMethod) //  this explicitly needs to named argument
}
```

Suggested changes:
- Add methods overloads that only take the block as one parameter so that you could write something like that:
```kotlin 
inState<S1> {
   onAction<A1> (::handlerMethod) //  no explicit named argument required
}
```

gabrielittner:
Closes #161 

This adds an `collectWhileInState` overload that `flowBuilder: (Flow<InputState>) -> Flow<T>` as parameter instead of a simple `Flow<T>`. The implementation is based on the same operator that onAction uses. This allows us to cancel it when the state changes without missing any actions while we are in the action. Any action emission is mapped into the current state and `distinctUntilChanged` is used to filter out emissions where the state did not change. The resulting `Flow` is then passed to `flowBuilder` and the return value of that is what we use like usually.

While I was adding the overload to `InStateBuilderBlock` I've also updated the docs on all methods, added overloads without `flatMapPolicy` (closes #179) and renamed `block` to `handler`.


gabrielittner:
The new operator crashes when running the native tests, I will look at that.

We can also see 2 different flakiness behaviors here:
- `on Action triggers only while in custom state` fails because one of the counters is 2 instead of 1, this looks like it might be an actual bug in the implementation or how the test is implemented
- 2 different tests failed once because they were already in the second state when we were expecting the initial state, I think this might be a test specific race condition. I'll check if my is assumption is right tomorrow.

dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/CollectStateWhileInStateTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlinx.coroutines.flow.flow
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.ExperimentalTime
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.flatMapConcat
+
+@OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class, ExperimentalTime::class)
+class CollectStateWhileInStateTest {
```
sockeqwe, line 14: Nitpicking: I'm wondering if we should merge this tests with the other  `CollectWhileInStateTest` to only have one "collectWhileInState" file with all tests of all variations...

OR

if we should have better naming conventions (i.e. class under test with "Test" suffix, i.e. then `CollectInStateSideEffectBuilderTest` ... 

I lean more towards the first option (merge all `inState<S> { collectWhileInState(...) }` into 1 file ...

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -24,51 +25,136 @@ class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
...
+     * An ongoing [handler] is cancelled when leaving this state or when a new [SubAction] is
+     * dispatched.
+     */
     inline fun <reified SubAction : A> on(
-        flatMapPolicy: FlatMapPolicy = FlatMapPolicy.LATEST,
-        noinline block: OnActionBlock<InputState, S, SubAction>
+        noinline handler: OnActionHandler<InputState, S, SubAction>
     ) {
+        on(FlatMapPolicy.LATEST, handler)
+    }
```
sockeqwe, line 39: 💪

sample/shared_code/src/commonMain/kotlin/com/freeletics/flowredux/sample/shared/PaginationStateMachine.kt
```diff
@@ -92,15 +92,15 @@ internal class InternalPaginationStateMachine(
...
-                onEnter(block = ::loadNextPage)
+                onEnter(::loadNextPage)
             }
 
             inState<ShowContentAndLoadingNextPageErrorPaginationState> {
-                onEnter(block = ::moveToContentStateAfter3Seconds)
+                onEnter(::moveToContentStateAfter3Seconds)
             }
         }
     }
```
sockeqwe, line 106: Much better to readability 🎉

sockeqwe (changes requested):

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -24,51 +25,136 @@ class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
...
+     *
+     * [handler] will only be called for a new emission from [flow] after a previous [handler]
+     * invocation completed.
+     */
+    fun <T> collectWhileInState(
+        flow: Flow<T>,
+        handler: InStateObserverHandler<T, InputState, S>
+    ) {
+        collectWhileInState(flow, FlatMapPolicy.CONCAT, handler)
+    }
```
sockeqwe, line 94: 💪

sockeqwe:
I could also think about a more specific use case  and therefore a convenience method would be nice where the orifinal state that triggered  `collectWhileInstate` is passed as a parameter but it should not invoke again again with newer state updates. 

Something of this method signature:

```kotlin
fun collectWhileInState(  
        flowBuilder: (InputState) -> Flow<T>, // so not a Flow<InputState>  but just `InputState` and only triggers once with the inputState that caused collectWhileInState to start
        handler: InStateObserverHandler<T, InputState, S>
)
```

i.e. if you want to observe a database for 1 particular Column / item by primary key, with SQLDelight or Room you don't want to cancel collection of query and start again because of flatMap transformation or so.

```kotlin
inState<ShowItemState> {
   collectWhileInState({ state : ShowItemState -> db.queryById(state.item.id) ) { item, state -> 
       MutateState { copy(item = item)  // So this triggers multiple times whenever the Item  in the database gets updated
    }
}
```

Another use case I could think of is connecting to a websocket  when you would like to start only once the websocket connection.


Under the hood it could be just a resuing what you are about to introduce with the PR:

```kotlin
val stateFlow : Flow<InputState> = ...

stateFlow.take(1).flatMapLatest { state -> flowBuilder(state) } ...
```

Do you think something like that could be useful? I'm happy to introduce something like this as a follow up PR ...

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectStateInStateSideEffectBuilder.kt
```diff
@@ -0,0 +1,60 @@
...
+    private val flowBuilder: (Flow<InputState>) -> Flow<T>,
+    private val flatMapPolicy: FlatMapPolicy,
+    private val handler: InStateObserverHandler<T, InputState, S>
+) : InStateSideEffectBuilder<InputState, S, A>() {
+
+    @Suppress("unchecked_cast")
+    override fun generateSideEffect(): SideEffect<S, Action<S, A>> {
+        return { actions: Flow<Action<S, A>>, getState: GetState<S> ->
+            actions.whileInState(isInState, getState) { inStateActions ->
+                inStateActions.map { getState() as InputState }
```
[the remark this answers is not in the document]
gabrielittner: > So while it works, I was wondering if instead of subscribing to the Actions directly collect the StateFlow from FlowReduxStateMachine?

We can't do that ourselves because we don't have the StateFlow available here. A user could do it with the existing `collectWhileInState`, but I think doing that is weird and you need to cast yourself.

Extracting parts of this into a function to make readability better makes sense

dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/CollectStateWhileInStateTest.kt
gabrielittner: I'm fine with merging 👍

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectStateInStateSideEffectBuilder.kt
```diff
@@ -0,0 +1,60 @@
+package com.freeletics.flowredux.dsl.internal
```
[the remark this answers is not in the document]
gabrielittner: How does `CollectInStateBasedOnStateSideEffectBuilder` sound?

dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/CollectStateWhileInStateTest.kt
gabrielittner: a3b7e45

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectStateInStateSideEffectBuilder.kt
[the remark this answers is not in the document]
gabrielittner: 9abaac9 this should improve the readability and also adds some comments to the extracted function.

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectStateInStateSideEffectBuilder.kt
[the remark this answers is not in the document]
sockeqwe: Sounds good to me 👍

sockeqwe (approved):
Thank you so much @gabrielittner  for working on this stuff 🙇‍♂️

gabrielittner:
> Have you considered making it an extension function on Flow<Action<S, A>>? For me, personally, having one parameter less (actions) improves readability a bit (not a game changer, thus up to you).

The reason why I didn't make it an extension is that I thought having a `currentStateFlow` method that gets some parameters makes more sense logically then transforming the actions Flow into a state Flow with an operator. If you just look at the point where it's called. From the implementation perspective it's still the same obviously. 

I don't have a strong opinion here, so feel free to change it.

> currentStateFlow(): In my head I still think about kotlin's StateFlow. No need to change, but maybe flowForCurrentState() or statemachinesStateAsFlow() ... Again, no need to change, can be merged as it is.

Updated to `flowOfCurrentState` 👍

sockeqwe:
👍 🚢

PR #183, opened 2021-08-04
make FlowReduxStateMachine lazy

gabrielittner:
This pr changes that the `FlowReduxStateMachine` does not immediately launch something when spec is called, which usually happens in the constructor. Instead it will just create a `StateFlow` there that has `SharingStarted.Lazily`. This makes it so that the collection of upstream only starts with the first subscriber. It's basically only becoming hot at that point.

Part 2: Don't use `FlowReduxStateMachine` in tests but instead a similar class that does not use `stateIn`. The reason for that is that `StateFlow` conflates values. If the state machine switches states immediately after we start the collection (even with the lazy changes mentioned above) if the  collector from turbine didn't get the initial state yet, then StateFlow will not emit it.

### DIFF
--- a/dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StateMachine.kt
+++ b/dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StateMachine.kt
@@ -1,24 +1,46 @@
 package com.freeletics.flowredux.dsl
 
-import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.launch
 
 @OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class)
 fun StateMachine(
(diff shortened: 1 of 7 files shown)


dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StateMachine.kt
```diff
@@ -1,24 +1,46 @@
...
-        CoroutineScope(Dispatchers.Default),
-        CommandLineLogger
-    ){
-        init {
-            spec(builderBlock)
+): TestStateMachine {
+    return TestStateMachine(builderBlock)
+}
+@OptIn(ExperimentalCoroutinesApi::class, FlowPreview::class)
+class TestStateMachine  constructor(
```
sockeqwe, line 22: It's a pity that we cannot use `FlowReduxStateMachine` ... I am afraid that in the long run we could encounter issues specific to `FlowReduxStateMachine` that are not tested in `TestStateMachine`.


With that said, what we most want to test is the DSL, so it should ok if we could then write some sort of other Unit test that would cover the `FlowReduxStateMachine` internal behavior. At the moment I have a hard time to imagine how we could test `FlowReduxStateMachine`, do you see how we could do this? I still end up writing an integration test that would cover `FlowReduxStateMachine` plus DSL ...

gabrielittner:
It resolves at least the last case mentioned there because that had immediately changing the state away from initial. `on Action triggers only while in custom state` might still have issues, not entirely sure about that one but also didn't have a deeper look.

dsl/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StateMachine.kt
gabrielittner: Testing `FlowReduxStateMachine` should be generally fine as long as the test doesn't do something like 
```
onEnter {
  OverrideState(OtherState)
}
```

sockeqwe (approved):

PR #188, opened 2021-08-07
simplify CollectInStateSideEffectBuilder implementation

gabrielittner:
It now uses the same implementation as `onEnter` to detect entering and leaving the state.

I think we could even go further and remove the extra builder and use the `CollectInStateBasedOnStateSideEffectBuilder` also for this one. The given `Flow` can be converted into a flowBuilder like this `{ currentStateFlow -> currentStateFlow.take(1).flatMapLatest { flowPassedInByUser }`.

### DIFF
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
@@ -12,6 +12,7 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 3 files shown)


dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
```diff
@@ -38,15 +39,15 @@ internal class CollectInStateBuilder<T, InputState : S, S : Any, A : Any>(
...
+                .map { isInState(getState()) }
+                .distinctUntilChanged()
+                .flatMapLatest {
+                    if (it) {
+                        flow.flatMapWithPolicy(flatMapPolicy) {
+                            setStateFlow(value = it, getState = getState)
                         }
-                        FilterState.StateChanged.UNSUBSCRIBE -> flow { }
+                    } else {
+                        flowOf()
```
hoc081098, line 50: IMO, it should be `emptyFlow()` (singleton object)

gabrielittner: Good point 👍 583f080

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
```diff
@@ -38,15 +39,15 @@ internal class CollectInStateBuilder<T, InputState : S, S : Any, A : Any>(
...
             actions
-                .filterState(getState = getState, isInState = isInState)
-                .flatMapLatest { stateSubscription ->
-                    when (stateSubscription) {
-                        FilterState.StateChanged.SUBSCRIBE -> {
-                            flow.flatMapWithPolicy(flatMapPolicy) {
-                                setStateFlow(value = it, getState = getState)
-                            }
+                .map { isInState(getState()) }
+                .distinctUntilChanged()
```
sockeqwe, line 43: Similarly to last Time with ˋflowOfCurrentState()ˋ: l would extract this into another function to increase readability...

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
```diff
@@ -12,6 +12,7 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOf
```
sockeqwe, line 15: I think this import should Change to emptyFlow()

sockeqwe (approved):
Good work🥳

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateSideEffectBuilder.kt
gabrielittner: Extracted it into an operator called `mapToIsInState`. Also using that in `onEnter`

PR #205, opened 2021-09-03
Fix InvalidMutabilityException

### LINKED ISSUE #204: InvalidMutabilityException on iOS
With Version 0.8.0 there's an `InvalidMutabilityException` when initializing a `FlowReduxStateMachine` on iOS:

```
Uncaught Kotlin exception: kotlin.native.concurrent.InvalidMutabilityException: mutation attempt of frozen MyStateMachine@106e9c8
```

Looks like it's introduced in [this change](https://github.com/freeletics/FlowRedux/pull/183/files):

https://github.com/freeletics/FlowRedux/blob/1ae6213f86ddd40f05358f95e0e0ce678245d1cc/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt#L24

ychescale9:
Fixes #204.

Tested locally and no longer crashes.


sockeqwe:
Thanks for the PR 👍  Code wise it looks good to me 💪 

I just don't fully understand why we need it. 

@gabrielittner  do you want to review it as well?

gabrielittner:
I'm also not entirely sure why we need it. Do we freeze the state machine?

ychescale9:
I haven't looked into the state machine implementation to understand exactly what happened but based on the exception it looks like the `internalState` is being assigned on a different thread than the thread that instantiates the `StateMachine` instance.

I'll try to find sometime to write a test to reproduce the issue.

sockeqwe:
@ychescale9 we will most likely move away from `StateFlow`. Therefore, this could become obsolete. I just wanted to give you a headsup before you put more time and effort into this.

See #210

ychescale9:
Thanks for the heads up. I was able to workaround this by passing in a `scope` with the main dispatcher on native (which is required by some other library anyway).

Moving away from StateFlow and not requiring passing in a `scope` sounds great 👍

PR #213, opened 2021-09-19
fix crash in flowBuilder based collectWhileInState

gabrielittner:
I can't reproduce it in tests but it happened a few times in a state machine that is very busy (multiple actions per second) after some time. I assume there are cases where the state changes right before the flow reaches map. Using a safe cast with mapNotNull here should be fine as any change state would no-op anyways because of later state checks.

In general this is probably another reason to rethink the internals and make state changes more of a primary thing instead of having it purely based on action emissions that happen afterwards

### DIFF
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateBasedOnStateSideEffectBuilder.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/CollectInStateBasedOnStateSideEffectBuilder.kt
@@ -11,7 +11,7 @@ import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flow
-import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.mapNotNull
 
 /**
  * A builder to create a [SideEffect] that observes a Flow<T> as long as the redux store is in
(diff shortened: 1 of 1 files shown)


sockeqwe (approved):
I'm not sure how `getState()` can ever return `null` but fine for me to add this safety net. Could you maybe add a little comment to that line why we use `mapNotNull`?

gabrielittner:
Generally the `whileInState` that we have should prevent the flow to run while we are not in the state and `getState() as InputState` should never result in a cast exception. However I've experienced it a few times that it failed with a `ClassCastException`. I think that's because `getState()` reaches into the store and cancellation of the flow isn't immediate. So I've changed the cast to a safe cast with `as?` and then `map` to `mapNotNull` to filter out the resulting nulls.

sockeqwe:
But if it runs into a `ClassCastException` when does `getState()` return `null`? 

Shouldn't the implementation then try to catch the `ClassCastException`? so something like this:

```kotlin
private fun flowOfCurrentState(actions: Flow<Action<S, A>>, getState: GetState<S>): Flow<InputState> {
        // after every state change there is a guaranteed action emission so we use this 
        // to get the current state
        return actions.mapNotNull { try {getState() as InputState} catch (e : ClassCastException) { null } }
            // an action emission does not guarantee that the state changed so we need to filter
            // out multiple emissions of identical state objects    
            .distinctUntilChanged()
    }
 ```

gabrielittner:
The `as?` with question mark is essentially doing that. It casts if possible and otherwise returns null

sockeqwe:
you are completely right 🙈  👍 

I am jumping too much between programming languages 😄  `as?` translated in my head to `as InputState?` ... My bad. All good, let's merge this 👍

DennisBlock:
Still get this issue. Got the `ClassCastException` with version 1.0.x when using the `flowBuilder` in `collectWhileInState`.

sockeqwe:
@DennisBlock any chance to post some code so that we can try to reproduce it somehow?

gabrielittner:
A full stack trace would also be helpful. Feel free to open a new issue with that

PR #217, opened 2021-10-03
Make FlowReduxStateMachine "cold" again

### LINKED ISSUE #204: InvalidMutabilityException on iOS

### LINKED ISSUE #210: Make FlowReduxStatemachine "cold" again
At the beginning FlowRedux the `FlowReduxStatemachine` was a "cold flow" (meaning on every new `state.collect()` the statemachine started with the initial state again).  

Starting in FlowRedux 0.6.0 we then have made `FlowReduxStatemachine` a "hot flow" and have started using `StateFlow` (hot means the same state machine state will be collected / shared by all collectors; so it doesn't start with the initial state for every new `state.collect {}` ; plus statemachine can run without any active `Collector` as long as the CoroutineScope where the state machine launched in is active).

We thought this could give us some advantages with Freeletic's DI setup and Jetpack Compose, however it turns out the drawback's are bigger than expected. Just a few pain points:
- Cancelation of a shared flow (like FlowReduxStateMachine) is harder
- Tests became more flaky because FlowReduxStateMachine starts working without active Collectors that have called `state.collect {...}`. That leads to some state transitions being missed in unit test because the collector used in the test was not subscribed yet.

Therefore we are going to revert this. We will move away from `StateFlow` and make FlowReduxStateMachine "cold" again .

sockeqwe:
Fixes #210 .

Removes `StateFlow` and the need to pass a `Scope` as constructor parameter of `FlowReduxStateMachine` in.

Also I have removed temporarily the inherintance to Freeletics MAD `StateMachine` because `StateMachine` still requires a `val state : StateFlow<S>` . Let's discuss how to proceed here, but I'm happy to create a PR to Freeletics MAD repo as well. 

It should also fix #204 .

### DIFF
--- a/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
+++ b/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
@@ -6,51 +6,66 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.channels.Channel
-import kotlinx.coroutines.flow.SharingStarted
-import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.consumeAsFlow
-import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.isActive
 
 @FlowPreview
 @ExperimentalCoroutinesApi
 abstract class FlowReduxStateMachine<S : Any, A : Any>(
-    private val initialState: S,
-    private val scope: CoroutineScope,
+    private val initialStateSupplier: () -> S,
     private val logger: FlowReduxLogger? = null
-) : StateMachine<S, A> {
+) {
 
     private val inputActions = Channel<A>()
-    private var internalState: StateFlow<S>? = null
+    private var specBlock: (FlowReduxStoreBuilder<S, A>.() -> Unit)? = null
+    private var specBlockSet = false
+
+    constructor(
+        initialStateSupplier: () -> S
+    ) : this(
+        logger = null,
+        initialStateSupplier = initialStateSupplier
+    )
+
+    constructor(initialState: S) : this(logger = null, initialState = initialState)
+
+    constructor(
+        initialState: S,
+        logger: FlowReduxLogger?
+    ) : this(logger = logger, initialStateSupplier = { initialState })
 
     protected fun spec(specBlock: FlowReduxStoreBuilder<S, A>.() -> Unit) {
-        if (internalState != null) {
+
+        if (this.specBlock != null)
             throw IllegalStateException(
                 "State machine spec has already been set. " +
-                    "It's only allowed to call spec {...} once."
+                        "It's only allowed to call spec {...} once."
             )
-        }
-        internalState = inputActions
-            .consumeAsFlow()
-            .reduxStore(logger, initialState, specBlock)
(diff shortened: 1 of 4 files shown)


dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -1,56 +1,69 @@
...
+        if (this.specBlock != null)
             throw IllegalStateException(
                 "State machine spec has already been set. " +
-                    "It's only allowed to call spec {...} once."
+                        "It's only allowed to call spec {...} once."
             )
-        }
-        internalState = inputActions
-            .consumeAsFlow()
-            .reduxStore(logger, initialState, specBlock)
```
gabrielittner, line 35: We could still have this here , just without the `stateIn`. The flow and therefore the state machine is still cold. Then we don't need `specBlock` and `specBlockSet` and cleanup for those is not needed

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -1,56 +1,69 @@
...
-    private var internalState: StateFlow<S>? = null
+    private var specBlock: (FlowReduxStoreBuilder<S, A>.() -> Unit)? = null
+    private var specBlockSet = false
+
+    constructor(
+        initialStateSupplier: () -> S
+    ) : this(
+        logger = null,
+        initialStateSupplier = initialStateSupplier
+    )
```
gabrielittner, line 26: Isn't this already covered by the default value on `logger`?

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -1,56 +1,69 @@
...
+
+    constructor(initialState: S) : this(logger = null, initialState = initialState)
+
+    constructor(
+        initialState: S,
+        logger: FlowReduxLogger?
+    ) : this(logger = logger, initialStateSupplier = { initialState })
```
gabrielittner, line 33: Can we just have the second one with `logger: FlowReduxLogger? = null`?

gabrielittner:
> Also I have removed temporarily the inherintance to Freeletics MAD StateMachine because StateMachine still requires a val state : StateFlow<S> . Let's discuss how to proceed here, but I'm happy to create a PR to Freeletics MAD repo as well.

Aside from changing the interface itself, the main thing that we need in MAD is to change the the whetstone codegen. It's currently generating `val state = stateMachine.state.collectAsState()` (example test https://github.com/freeletics/mad/blob/main/whetstone/compiler/src/test/kotlin/com/freeletics/mad/whetstone/codegen/FileGeneratorTestCompose.kt#L129) which wouldn't compile anymore because for a regular `Flow` it would require passing a default value. So we can either create our own `Flow<T>.collectAsState()` extension function an use that in the codegen (just change the import) or generate what such a function would do.

PR #226, opened 2021-10-25
Add test to check if flow is cold and dispatching only works to last cold flow

sockeqwe:
apparently dispatching DOES dispatch to all active flows (see failing test).

... not exactly what I was expecting ... but I guess `receiveAsFlow()` works differently that I was expecting ... 

btw. do we need a way to close the `Channel` or when will it get closed?


sockeqwe:
This is still an issue but I'm not sure if it is actually a bug or a feature 😄 

So the problem is: We want to have `FlowReduxStateMachine.state` as a cold flow. That is ture at the moment: everytime you call `FlowReduxStateMachine.state.collect { ... }` we start "cold" with the initial state.'

Given the internal implementation details, however, multiple cold flows share the same `Channel<Action>` under the hood that effectively dispatches the same action to all still active flows (created previously by calling `FlowReduxStateMachine.state.collect { ... }`) 

See implementation:

https://github.com/freeletics/FlowRedux/blob/28701a440a8bd303199d3a1d8d364b82cb104c95/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt#L36-L44

With


https://github.com/freeletics/FlowRedux/blob/28701a440a8bd303199d3a1d8d364b82cb104c95/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt#L21-L22

and `dispatch()` using the same single instance of `Channel`

https://github.com/freeletics/FlowRedux/blob/28701a440a8bd303199d3a1d8d364b82cb104c95/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt#L53-L63

So the result is: 

```kotlin
val myStateMachine = FlowReduxStateMachine<S,A>()

launch {
    // Job 1
   myStateMachine.state.collect { ... }  // start cold
   
   delay(100)
   
   myStateMachine.dispatch(Action1) // this action will be dispatched also to Flow in job2 (see below)
}

launch {
    // Job 2
   myStateMachine.state.collect { ... }  // start cold
}
``` 


So the  solution could be to  change implementation details to create a new channel everytime  `FlowReduxStateMachine.state` is called  to not have this interfearing of dispatching Actions in a "shared way".


However, I am not sure if this is actually an issue or more a feature 🙈  

Thoughts @gabrielittner ?

gabrielittner:
What if we would use the counter that we have to disallow dispatching actions before someone starts collecting the flow to also disallow collecting the same state machines more than once at a time?

sockeqwe:
That is an option. I actually like it as I think it gives some kind of "predictability" how things work.

PR #230, opened 2021-11-14
enable explicitApi, cleanup exposed internal code

gabrielittner:
The reified inline methods now have a non inline overload that takes the class as parameter. This way we can completely hide the internal lists of builders


dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/StoreWideBuilderBlock.kt
```diff
@@ -13,7 +13,7 @@ import com.freeletics.flowredux.dsl.internal.Action
  * That is not that nice as it all subclasses are implementation detail heavy.
  * There is no need to have a hundreds of lines of code in one file just to have sealed classes.
  */
-abstract class StoreWideBuilderBlock<S, A> internal constructor() {
+public abstract class StoreWideBuilderBlock<S, A> internal constructor() {
```
gabrielittner, line 16: Should we remove the base class as long as it only has one usage?

sockeqwe: Yea probably 👍

sockeqwe:
General question: 
don't you get now at least 2 autocomplete options for in the IDE? One fore `inState<Foo> { ... }` one for `inState(Foo::class) { ... } `?

I would prefer to not have these autocomplete overhead ...

gabrielittner:
Maybe we can give those methods a different name? So that it at least doesn't appear when you type the correct first letter.

sockeqwe:
Actual, is it so bad to expose
`lists of builders` ? Then people could expand DSL if they need to with their own functionality (more than just reusing existing DSL blocks)

gabrielittner:
It's also requring the `OnActionInStateSideEffectBuilder` to be public, not just the builders list. If it were just that (or if we want to promote extensibility) I'd add a public`addBuilder` method

sockeqwe (approved):
👍

Issue #139, opened 2021-06-19
Artifact for Jetpack Compose


sockeqwe:
We could provide a convinience artefact for jetpack compose that
- picks right coroutinescope to subscribe on, i.e.
   ```kotlin
     @Composable 
     fun MyComponent() {
          val (state : State<MyState>, dispatch : (MyAction) -> Unit)  by stateMachine(MyFlowReduxStateMachine) // Will do remember { mutableStateOf( ... ) } under the hood and start obsering the StateMachine.state property on the right coroutine scope
     }
   ```
- would also allow to use FlowRedux "inline" i.e. to build state for a combination of UI widgets only 
```kotlin
     @Composable 
     fun MyComponent() {
          val (state : State<MyState>, dispatch : (MyAction) -> Unit)  by stateMachine { 
             // statemachine spec DSL
                inState<S1> {
                   onEnter { ... }
                   onAction { ... }
                }  
                ...
           }
     }
   ```

sockeqwe:
Has been released with 0.11.0

PR #236, opened 2021-11-19
Add an convenient Artifact to use `FlowReduxStateMachine` in compose

sockeqwe:
Adds a way to subscribe to state in compose:
- `MyFlowReduxStateMachine.asState()` 

and a way to get both, subscribe to state and dispatch actions:
- `val (state : State<MyState>, dispatch : (MyAction) -> Unit) = MyFlowReduxStateMachine.stateAndDispatch()`

I have also added a compose code sample. For sure it can be made nicer or more idiomatic, but does the job so far I guess.


dependencies.gradle
```diff
@@ -1,9 +1,9 @@
 ext.versions = [
         agp                    : '7.0.3',
-        compileSdk             : 30,
-        targetSdk              : 30,
+        compileSdk             : 31,
+        targetSdk              : 31,
```
sockeqwe, line 4: It also updates compile + target SDK to `31`(then some other DependABot created PRs should compile)

sample/android/src/main/java/com/freeletics/flowredux/compose/PopularRepositoriesUi.kt
```diff
@@ -0,0 +1,62 @@
...
+                is LoadingFirstPageError -> ErrorUi(dispatch)
+                is ContainsContentPaginationState -> {
+                    val showLoadNextPageUi = s.shouldShowLoadMoreIndicator()
+                    val showErrorSnackBar = s.shouldShowErrorSnackbar()
+
+                    ReposListUi(repos = s.items, loadMore = showLoadNextPageUi, dispatch = dispatch)
+
+                    val errorMessage = stringResource(R.string.unexpected_error)
+                    if (showErrorSnackBar) {
+                        LaunchedEffect(scaffoldState.snackbarHostState) {
```
sockeqwe, line 39: I'm actually not 100% sure if this works as intended because of the surrounding `if (showErrorSnackBar)` ... It seems to work, but here I don't understand how `LaunchedEffect(key)` learns that the `key = scaffoldState.snackbarHostState` has changed if it surrounded by an `if(showErrorSnackBar)`  ...

sample/android/src/main/java/com/freeletics/flowredux/compose/ReposListUi.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        if (loadMore) {
+            item {
+                LoadNextPageUi()
+            }
+        }
+    }
+
+    if (loadMore) {
+        LaunchedEffect(loadMore) {
```
sockeqwe, line 42: Same here: https://github.com/freeletics/FlowRedux/pull/236/files#r752936643

sample/shared_code/src/commonMain/kotlin/com/freeletics/flowredux/sample/shared/GithubApi.kt
```diff
@@ -8,25 +8,27 @@ class GithubApi {
     // Used to simulate network errors
     private var counter = 0
     private fun shouldFail(): Boolean =
-            counter++ % 4 == 0
+        counter++ % 4 == 0
 
     suspend fun loadPage(page: Int): PageResult {
-        delay(300)
+        delay(2000)
```
sockeqwe, line 14: increased to see things better in the Sample Application UI

sample/shared_code/src/commonMain/kotlin/com/freeletics/flowredux/sample/shared/GithubApi.kt
```diff
@@ -8,25 +8,27 @@ class GithubApi {
...
         val end = min(githubData.size, page * pageSize + pageSize)
 
         return (
-                if (start < githubData.size) githubData.subList(start, end) else emptyList<GithubRepository>()
+                if (start < githubData.size) githubData.subList(
+                    start,
+                    end
+                ) else emptyList<GithubRepository>()
                 ).run {
-                    // TODO rewrite this to use header Link
```
sockeqwe, line 23: Will never happen 😄  therefore removed this comment 😺

sample/shared_code/src/commonMain/kotlin/com/freeletics/flowredux/sample/shared/PaginationStateMachine.kt
```diff
@@ -73,7 +73,14 @@ data class ShowContentAndLoadingNextPageErrorPaginationState(
...
 
-internal class InternalPaginationStateMachine(
+/**
+ * It's callend `Internal` because it is note meant to be accessed publicly as it exposes coroutines
+ * Flow and suspending function to dispatch.
+ *
+ * Instead the "wrappter class" [PaginationStateMachine] should be used which hides `Flow` etc.
+ * but uses traditional "callbacks". That way it is easier to use on iOS.
+ */
+class InternalPaginationStateMachine(
```
sockeqwe, line 83: The idea was to only expose publicly `PaginationStateMachine` and not this class `InternalPaginationStateMachine` because `PaginationStateMachine` provides traditional callback based listener so that from iOS you don't need to bridge `Flow` and `suspend function`. 

but compose actually should access directly a `FlowReduxStateMachine` to be able to use extension functions like `stateAndDispatch()` on it. Therefore made it public.

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -15,10 +15,12 @@ import kotlinx.coroutines.sync.withLock
 @FlowPreview
 @ExperimentalCoroutinesApi
 abstract class FlowReduxStateMachine<S : Any, A : Any>(
-    private var initialStateSupplier: () -> S,
+    initialStateSupplier: () -> S,
     private val logger: FlowReduxLogger? = null
 ) : StateMachine<S, A> {
 
+    val initialState: S by lazy(LazyThreadSafetyMode.NONE, initialStateSupplier)
```
sockeqwe, line 22: So initial state will be kept forever. That ensures that it is always the same initial state for all `state:Flow<MyState>` collectors. Also with making this public, `FlowReduxStateMachine.asState()` Compose extension can access it and return type is `State<MyState>` instead of nullable  `State<MyState?>` as it serves as the initial state in `produceState {...}`.

sample/android/src/main/java/com/freeletics/flowredux/compose/PopularRepositoriesUi.kt
```diff
@@ -0,0 +1,62 @@
...
+@Composable
+fun PopularRepositoriesUi() {
+    val stateMachine = InternalPaginationStateMachine(
+        githubApi = GithubApi(),
+        logger = AndroidFlowReduxLogger
+    )
+
+    val (state, dispatch) = stateMachine.stateAndDispatch()
+
+    // Timber.d("New State ${state.value}")
```
sockeqwe, line 23: @gabrielittner  here I run in an interesting issue. If I let this line execute, then on dispatching an action I get the following error message:


> java.lang.IllegalStateException: Cannot dispatch action com.freeletics.flowredux.sample.shared.RetryLoadingFirstPage@325bf06 because state Flow of this FlowReduxStateMachine is not collected yet. Start collecting the state Flow before dispatching any action.
>         at com.freeletics.flowredux.dsl.FlowReduxStateMachine.dispatch$suspendImpl(FlowReduxStateMachine.kt:67)
>         at com.freeletics.flowredux.dsl.FlowReduxStateMachine.dispatch(Unknown Source:0)
>         at com.freeletics.flowredux.compose.ComposeFlowReduxStateMachineKt$stateAndDispatch$1$1.invokeSuspend(ComposeFlowReduxStateMachine.kt:54)
>         at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
>         at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
>         at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:81)
>         at androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
>         at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.run(AndroidUiDispatcher.android.kt:57)
>         at android.os.Handler.handleCallback(Handler.java:883)
>         at android.os.Handler.dispatchMessage(Handler.java:100)
>         at android.os.Looper.loop(Looper.java:214)
>         at android.app.ActivityThread.main(ActivityThread.java:7356)
>         at java.lang.reflect.Method.invoke(Native Method)
>         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
>         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

Obviously this is coming from one of our internal checks
https://github.com/freeletics/FlowRedux/blob/088ef8a4d5d78e81c015209c0df96cd9cc764914/dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt#L65

But what it should mean is that the previously collected `Flow` is not collected anymore. That doesn't make any sense to me. Do you see how this is caused or where it is coming from?

sockeqwe:
Closes #139

sample/android/src/main/java/com/freeletics/flowredux/compose/PopularRepositoriesUi.kt
sockeqwe: Nevermind, I figured it out. Just ignore my comment above.

PR #263, opened 2022-02-16
Add usage of atomic variables

BenjaminLifeTime:
When working in a kotlin multiplatform environment and using this state machine the following problem occurs:
If activeFlowCounter is not atomic an InvalidMutabilityException is thrown on iOS devices as soon as the state machine runs, i.e. a state is changed. Due to the way memory is managed there the variable in question is frozen and must not be mutated. Having an atomic int helps to solve this issue.


BenjaminLifeTime:
Thanks for the reply.
As I share your attitude towards dependencies I have already removed it and added everything that is needed to solve this issue.
I also hope that kotlin's new memory management model will solve a lot of problems like these, but you can never be sure and this solution should work either way.
To reproduce the crash I've created a little demo project, which you can find here: git@github.com:BenjaminLifeTime/FlowReduxDemo.git
Just remove the lines 20-25 from settings.gradle.kts to make it use the current release of FlowRedux (v0.11.0) and run the iOS app (Android won't crash). To test the fix just add these lines again, have the FlowRedux project in a parallel folder and checkout the 'feature/atomics_usage' branch.

dependencies.gradle
```diff
@@ -28,6 +28,7 @@ ext.versions = [
 
         junit                  : '4.13.2',
         turbine                : '0.7.0',
+        stately                : '1.2.0'
```
hoc081098, line 31: Is this unused?

BenjaminLifeTime: What the...?!? Sorry about that, must have snuck in there when rebasing and rearranging my git history.

sockeqwe:
Thanks for your PR 🙇  Overall looks good 👍 

I am still wondering where this can actually be exactly an issue. Do you call `FlowReduxStateMachine.state.collect { ... }` from different threads?

BenjaminLifeTime:
Thanks!😁
Overall, iOS seems to be extremely touchy when it comes to changing properties in coroutines, which has been the issue here, too. I really hope that the upcoming memory management overhaul will bring a change for the better.

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
```diff
@@ -0,0 +1,21 @@
+package com.freeletics.flowredux.dsl.util
+
+import kotlin.native.concurrent.AtomicInt
+
+public actual class AtomicInt actual constructor(initialValue: Int) {
+    private val atom = AtomicInt(initialValue)
```
sockeqwe, line 6: should this use full package name to avoid resolving to the "wrong" ˋAtomicIntˋ?

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2022 LifeTime GmbH. All rights reserved.
+ */
+
+package com.freeletics.flowredux.dsl.util
+
+public expect class AtomicInt(initialValue: Int) {
```
sockeqwe, line 7: Since we only use  `get()`, `increment()` and `decrement()` I would like to propose to limit the publi API only to this methods and maybe even rename the class to `AtomicCounter` to be more explicit about what this class will be about.

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
```diff
@@ -0,0 +1,21 @@
+package com.freeletics.flowredux.dsl.util
+
+import kotlin.native.concurrent.AtomicInt
+
```
sockeqwe, line 4: Do you think it makes sense to write some unit tests for this class? In particular multithreaded ones?

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
sockeqwe: > 

Comment above could be obsolete if we rename class to AtomiCounter as proposed in another comment above

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
hoc081098, line 7: How about `internal expect class`?

sockeqwe: 💯

sockeqwe: And we could make this `internal`

BenjaminLifeTime: Unfortunately, this class cannot be made internal due to the `typealias`. `AtomicInteger` is public and thus `AtomicCounter` has to be public, too, I'm afraid.

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
BenjaminLifeTime: As we only have a wrapper and a typealias here tests are redundant in my opinion. We would only test ` kotlin.native.concurrent.AtomicInt` and `java.util.concurrent.atomic.AtomicInteger`, which should have already been tested thoroughly (, at least I hope so 😉)

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
BenjaminLifeTime: Redundant methods are removed. I'd love to make it `internal` but this isn't possible due to the typealias in jvmMain.

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
BenjaminLifeTime: 👍🏻Renamed

dsl/src/nativeMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicCounter.kt
```diff
@@ -0,0 +1,17 @@
+package com.freeletics.flowredux.dsl.util
+
+import kotlin.native.concurrent.AtomicInt
+
+public actual class AtomicCounter actual constructor(initialValue: Int) {
+    private val atom = AtomicInt(initialValue)
+
+    public actual fun get(): Int = atom.value
+
+    public actual fun set(newValue: Int) {
```
sockeqwe, line 10: Nitpicking, but I also think `set` also doesn't need to be public API as we dont need it right now (just get and increment, decrement). But not a blocker, we can clean that up later imho
I

dsl/src/commonMain/kotlin/com/freeletics/flowredux/dsl/util/AtomicInt.kt
sockeqwe: We could in a follow up discussion talk if it makes sense to wrap jvm's `AtomicInt` around `AtomicCountet` (same way you did it for native targets so that we could mark everything as `internal`). But for now not a blocker!

sockeqwe:
thank you!

PR #265, opened 2022-02-23
publish everything from macOS

gabrielittner:
With Kotlin 1.6.0 it's possible to build the windows artifacts on macOS so we can build everything there for publishing.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -7,14 +7,7 @@ on:
 
 jobs:
   publish:
-
-    strategy:
-      matrix:
-        os: [macos-latest, windows-latest]
-      # run sequentially to avoid conflicts for closeAndReleaseRepository
-      max-parallel: 1
-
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -27,20 +20,12 @@ jobs:
 
       - name: Install Android SDK
         run: ./.github/android-sdk.sh
-        if: matrix.os == 'macos-latest'
+
+      - name: Build release
+        run: ./gradlew assemble
```
gabrielittner, line 25: Adding this as an extra step. I've found that building and publishing in one step can lead to having to staging repos on sonatype if the build is slow (it builds one artifacts, publishes it, builds another, a lot of time has passed, publishes the second artfact to a new repo). This then leads to `closeAndReleaseRepository` to fail. This way the following `publish` can just use the already built binaries

sockeqwe (approved):

Issue #270, opened 2022-03-11
Change the DSL artifact to be the main artifact, make reduxStore internal


gabrielittner:
I've been wondering whether "FlowRedux" should just be the DSL artifact and the `reduxStore` is just an internal implementation detail. That would also allow us significantly change the implementation later on (like #209) without having to maintain the `reduxStore` method.

sockeqwe:
Agreed, no objections from my side to change that!

Gabriel Ittner ***@***.***> schrieb am Fr., 11. März 2022,
13:32:

PR #274, opened 2022-03-20
Remove DLS module and move things into1 single flowredux module

sockeqwe:
Addresses #270 : Merges `dsl` module into `flowredux` module so that only 1 modules lives along with the `compose` module.

Open points:
- [x] Merge code (incl. tests) from dsl module into flowredux module
- [x] make internal API internal (modifiers)
- [x] Remove duplicated code such as `TestUtils` etc.
- [x] update README

Open question: Do we also want to rename packages? I think keeping things under `flowredux.dsl` is not too bad to have the possibility to add things somewhen in later versions directly under `flowredux` package. Also doesn't break backward compatibility.


gabrielittner:
I think keeping the package is fine 👍

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/SideEffect.kt
```diff
@@ -11,10 +11,10 @@ import kotlinx.coroutines.flow.Flow
  * @param state [GetState] to get the latest state of the state machine
  */
 // TODO find better name?
```
gabrielittner, line 13: Should we also remove this todo?

sockeqwe:
@gabrielittner  ready for review now. Could you please take a look? Maybe I have missed something

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/SideEffect.kt
sockeqwe: good catch 👍  

Removed

gabrielittner (approved):
I didn't find anything that's left 👍

sockeqwe:
I will create a new release now with these changes

PR #276, opened 2022-03-21
remove StoreWideBuilderBlock

gabrielittner:
Follow up to https://github.com/freeletics/FlowRedux/pull/230#discussion_r748865885 Since we only have one `StoreWideBuilderBlock` and no plans of introducing others we can remove it for now.


sockeqwe (approved):
👍  makes sense

PR #279, opened 2022-03-21
introduce currentState in FlowReduxStateMachine

gabrielittner:
When the compose artifact was introduced we started keeping `initialState` in an internal variable inside of `FlowReduxStateMachine` even if we passed in an `initialStateSupplier`. This has the issue that whenever you start collecting the state machine again you will start from what `initialStateSupplier` returned when you first started collecting. This means that you can't for example safe the state and then restore it in the supplier. It doesn't just affect configuration changes, it also happens for example if you use Fragments navigate to another one and then go back. 

What this does is that it introduces an internal `currentState` variable that is updated from an `onEach`. Whenever we build a state flow we will use `currentState` if available and only fallback to `initialState` if it's not. There is a new `stateAndInitialState` method that returns the `state` together with what will be the initial state of that `Flow` (modeled after `StateAndDispatch`). I didn't want to make `currentState` itself public because it would enable bad state machine implementations that try to read it.

This has one effect on child state machines. If you re-use the same instance of a child state machine when leaving and re-entering that child state it will behave differently because it still has the old `currentState`. If the child state machine left the state it's supposed to operate in, it's effectively dead because it will never re-enter the state. You can see that in the updated test, the child state machine will forever stay in `S1` after the first loop. Personally I think this change is fine because you'd probably construct a new sub state machine each time and pass the parent state to it, but maybe there is a use case I'm not considering.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.receiveAsFlow
 
@@ -16,7 +17,8 @@ public abstract class FlowReduxStateMachine<S : Any, A : Any>(
     initialStateSupplier: () -> S,
 ) : StateMachine<S, A> {
 
-    public val initialState: S by lazy(LazyThreadSafetyMode.NONE, initialStateSupplier)
+    private val initialState: S by lazy(LazyThreadSafetyMode.NONE, initialStateSupplier)
+    private var currentState: S? = null
 
     private val inputActions = Channel<A>()
     private lateinit var outputState: Flow<S>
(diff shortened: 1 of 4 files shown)


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -50,6 +53,10 @@ public abstract class FlowReduxStateMachine<S : Any, A : Any>(
             return outputState
         }
 
+    public fun stateAndInitialState(): StateWithInitialState<S> {
+        return StateWithInitialState(state, currentState ?: initialState)
+    }
```
gabrielittner, line 58: We could also have our one `@OptIn` annotation to mark this as internal so that it's only used from the compose artifact.

sockeqwe:
My thoughts:
Maybe it was a mistake to introduce the `initialState` field, maybe we should undo this.

The only reason why I have added `initialState` is because `produceState()` needs an initial state as parameter OR work with `null` as initial state (or introduce a sealed class modelling the absence of state yet at this point in time).

https://github.com/freeletics/FlowRedux/blob/cafdce4ac8c5a16764e3df8be53c4440e49c769d/compose/src/main/java/com/freeletics/flowredux/compose/ComposeFlowReduxStateMachine.kt#L18-L22

```kotlin
public fun <S : Any, A : Any> FlowReduxStateMachine<S, A>.rememberState(): State<S> {
    return produceState(initialValue = this.initialState, this) {
        state.drop(1) // skip the first one as it is the initial state which is already submitted with produceState's initial state
            .collect { value = it }
    }
}

```

@gabrielittner  so what if we just go the other route and undo the `initialState` thing and we will find another solution for compose?

gabrielittner:
I'm fine with that. Internally we we use `null` as default state and don't render anything for null because we just make the assumption that the state machine will emit immediately. So far I haven't seen any issues with that approach.

Imo there is still a bit of value of remembering the last state but when the `initialStateSupplier` is invoked whenever you start collecting you can build that on top of `FlowReduxStateMachine`

PR #282, opened 2022-03-22
Removed initial state and use `null` as initial state in compose

sockeqwe:
As discussed in #279 :
- removing initial state from state machine. Maybe we can find some middle ground in the future between #279 and this PR but I think for now it is best to make it "predictable" and simplified by removing this.
- Uses `null` as initial state for `produceState() ` i.e. used by `rememberState()` and `rememberStateAndDispatch()`. I have mixed feelings about it. On one hand I don't like working with `null` too much and would prefer usually something more descriptive, but on the other hand introducing some linke sealed interface just to be more descriptive seems like an overkill to me. I played around in my head with something like this: 
```kotlin
public sealed interface FlowReduxState<S : Any>

/**
 * Indicates that no state has been emitted by the [FlowReduxStateMachine] yet.
 */
public object NoStateYet : FlowReduxState<Nothing>

@JvmInline
public value class CollectedState<S : Any>(private val internalState : Any) : FlowReduxState<S>{
    public fun  state() : S = internalState as S
}
```

(Btw. apparently Generics don't work directly with inline classes that is why I worked around it with `fun state()` and casting that made it even worse 😄  for the consumer of this library). 

Unless someone sees an issue, I would go for now with `null`.

### DIFF
--- a/compose/src/main/java/com/freeletics/flowredux/compose/ComposeFlowReduxStateMachine.kt
+++ b/compose/src/main/java/com/freeletics/flowredux/compose/ComposeFlowReduxStateMachine.kt
@@ -10,14 +10,22 @@ import kotlinx.coroutines.launch
 
 /**
  * Get a Compose [State] object from a [FlowReduxStateMachine].
+ *
+ * The returned [State.value] can be `null`.
+ * `null` means that the [FlowReduxStateMachine] has not emitted a state yet.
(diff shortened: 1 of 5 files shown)


gabrielittner (approved):
Thanks for making the change, I was busy with other things yesterday.

I think the internal `reduxStore` overload that takes `initialState` instead of the supplier might be unused now

sockeqwe:
You are right, will remove 👍

PR #306, opened 2022-06-03
replace state snapshot with new State class

### LINKED ISSUE #211: MutateState needs annoyingly Type Arguments
As a user of FlowRedux I need to specify explicitly the Type argiments of `MutateState`. Example:

```kotlin
sealed interface State

data class S1 (val i : Int) : State
object S2 : State
```

```kotlin
inState<S1> {
   onEnter { stateSnapshot ->
      ...
      MutateState<S1, State> { this.copy(i = this.i++) }  // MutateState needs S1 as type parameter; compiler cannot infer it.
   }
}
```

This is annoying and counterproductive to our goal of increase developer experience but I'm not sure how to solve this problem best. Some random ideas:

1. Since all blocks inside the `inState` block already know of which type the State we want to mutate is, we could add a extension function to the `onEnterBlock`, `onAction` and `collectWhileInstate` that would allow us to write something like this:
```kotlin
inState<S1> {
   onEnter { stateSnapshot ->
      ...
      mutateStateExtensionFunction { this.copy(i = this.i++) }  // this == S1 can be inferred by the compiler via extensionFuntion
   }
}
```
(Please note that `mutateStateExtensionFunction` is just a random name to illustrate the point I try to make; not the final name in case we go for this solution)
While this approach works for blocks directly defined in the DSL it would not work for functional references like this:
```kotlin
inState<S1> {
   onEnter(::doSomethingOnEnterS1)
}


fun doSomethingOnEnterS1(stateSnapshot: S1) : ChangeState<State> {
   ...
   return  stateSnapshot ->
      ...
      mutateStateExtensionFunction { this.copy(i = this.i++) }  // wont work because no such extension function defined for an aribtarry function
   }
}
```

So I was looking if it is possible to define extension functions on "Functions" but didn't figure it out yet how to do it (nor am I sure if the kotlin IDE plugin would work properly). The following did not work (but I didn't dive deeper):

```kotlin
fun <R> kotlin.Function<R>.mutateStateExtensionFunction() {...}

fun <InputState, S> ((stateSnapshot : InputState ) -> S).mutateStateExtensionFunction() {...}
```

Maybe #189  could introduce a new supertype we could define a `mutateStateExtensionFunction`  extension function on, but it doesn't look to solve the problem of function reference.

2. Another option is to go away from defining `MutateState` with receiver type.  Receiver type is handy especially with data classes used for State definition because then you can just call `copy()` to do a mutation as `this` is introduced by the compiler via `reciever type`. However, this introduces the problem that we need to tell in `MutateState` what the type of the receiver is causing developers to Explicitly specify the Generic Type Arguments.
If we would remove the receiver type, the compiler's Type inference should be smart enough to infer types (although we still need to give the compiler a hint somewhere, i.e. define the parameter type of the state snapshot):

```kotlin
inState<S1> {
   onEnter { stateSnapshot : S1 ->
      ...
      MutateState { currentState : S1 -> currentState .copy(i = currentState.i++) }  //  Explicity have defined input parameter currentState : S1  is enough for compiler to be satisfied
   }
}
```

this however doesn't fully solve the problem of developer's productivity but could be some compromise.

Do you see any other options? Do I miss something?

### LINKED ISSUE #288: Add more overloads for the InStateBlockBuilder methods
Currently when using method references you that the Kotlin compiler will flag your parameters as unused when you just add them to be to use method references. The most common case is probably an unused `stateSnapshot`, but for the action handlers it could even be the action itself if it's just an `object`. If you don't use method references this issue still exists. While you can just use `_` to get rid of the warning you still of the verbosity of having to write out the lambda parameters (`on<SomeAction> { action, _ ->`). 

Side note that influences potential solutions: I'm getting more and more convinced that we should not use method references anymore and rely on lambdas. The general aspiration of having the `spec` block as small as possible is still valid, however I think method references are not the right solution. There are multiple issues and while some of them are not our fault and could be fixed by Kotlin they still matter:
- the initially mentioned unused parameter issue which either means you get tons of compiler warnings or need to put suppress annotations everywhere
- in some circumstances (not sure when it happens) the IDE can't really auto complete the referenced function when you're writing `::...`
- the function name should describe what it does but in practice I'm usually seeing `on(::buttonClicked)`, my assumption is that if you have `on<ButtonClicked> { ... }` you are less inclined to call the method the same
- An addition to the above is that if you do `on(::transitionToSomeState)` you loose the information of when it happens while `on<ButtonClicked> { transitionToSomeState(it) }` would make getting an overview easier. Of course you can combine both in your function name but that leads to long/weird names. For example `onButtonClickedTransitionToSomeState` is long and would then result in `on(::on...)` while leaving out the `on` from the method name reads weirdly when looking at just the function
- Method references don't work well when the method has a dependency. In practice the function is just put into the state machine class which means it's not testable on it's own anymore. While we introduced functional interfaces in #189 to make this easier, they are also a lot more verbose. If you just call a method from the lambda you can pass additional parameters to it and it keeps the syntax more aligned.

Going back to the original problem depending on whether we want to encourage method references or not all [handlers](https://github.com/freeletics/FlowRedux/blob/main/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/Handlers.kt) would need overloads. If we are saying method references are the way to go then we'd need additional interfaces + overloads with 0 and 1 parameters for everything that has 1+ parameters. If we decide for lambdas then it would be interfaces + overloads for everything with 2+ parameters because a single parameter is just an implicit `it` either way.

gabrielittner:
For context see discussion in #211

This introduces a new `State<InputState>` class that gives access to the snapshot and has `mutate`, `override` and `noChange` methods to create `ChangeState` objects. This new class replaces the snapshot parameter on all non effect methods.

Advantages:
- it knows about the `InputState` type so it's not possible anymore to accidentally create `MutateState` instances with the wrong input type, this is now enforced by the type system which it wasn't before
- it's also not necessary anymore to specify the type parameters for mutate state, the generics are inferred

We also decision to prefer one line lambdas over method references (e.g. `on<ButtonClicked> { action, state -> transitionToSomeState(action, state) } over `on(::buttonClicked`/`on(::transitionToSomeState)`) because:
- Parameters to of the reference are required to be there but the compiler warns when they are unused, in a lambda you can just use `_` and remove the parameter from the method
- The function name should describe what it does (`doSomething`) but in practice I'm usually seeing in named after the action (e.g. `on(::buttonClicked)`), the assumption is that if you have `on<ButtonClicked> { ... }` you are less inclined to call the method `buttonClicked` as well
- An addition to the above is that if you do `on(::doSomething)` you loose the information of when it happens while `on<ButtonClicked> { transitionToSomeState(it) }` would make getting an overview easier. Of course you can combine both in your function name but that leads to long/weird names. For example `onButtonClickedTransitionToSomeState` is long and would then result in `on(::on...)` while leaving out the `on` from the method name reads weirdly when looking at just the function
- Method references don't work well when the method has a dependency. In practice the function is just put into the state machine class which means it's not testable on its own anymore. While we introduced functional interfaces in #189 to make this easier, they are also a lot more verbose. If you just call a method from the lambda you can pass additional parameters to it and it keeps the syntax more aligned.
- in some circumstances (not sure when it happens) the IDE can't really auto complete the referenced function when you're writing `::...`

Because of that and because I had to touch most of the DSL anyways I also made these changes
- remove functional interfaces
- remove the overloads without `ExecutionPolicy` and bring back default values for it (we introduced this to make method references easier to use in #179

Also renamed `stateMachine` to `onEnterStartStateMachine` to match the new `onActionStartStateMachine`.

Closes #211, closes #288


gabrielittner:
The snapshot publishing failed because of dokka, going to revert the update

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
```diff
@@ -21,12 +21,17 @@ public sealed class ChangeState<out S>
 /**
  * Sets a new state by directly override any previous state
  */
+@Deprecated("call override() on State instead", level = DeprecationLevel.WARNING)
 public data class OverrideState<S>(internal val newState: S) : ChangeState<S>()
```
gabrielittner, line 25: Kept these public as deprecated for now to make the migration at least a bit easier

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
```diff
@@ -35,16 +40,30 @@ public class MutateState<InputState : S, S>(
...
 /**
  * No change, this is semantically equivalent to use [OverrideState] and pass in the previous state
  */
+@Deprecated("call noChange() on State instead", level = DeprecationLevel.WARNING)
 public object NoStateChange : ChangeState<Nothing>()
 
+//TODO rename after removing deprecated NoStateChange
+internal object InternalNoStateChange : ChangeState<Nothing>()
+
 internal fun <S> ChangeState<S>.reduce(state: S): S {
```
sockeqwe, line 60: Not 100% sure, but if we make all these `ChangeState` implementation internal only then we cannot write simple unit tests for the "handlers" anymore, do we?

example

```kotlin
fun handleOnEnter(state : State<MyState>) : ChangeState<MyState> { ... } // I want to unit test this function

@Test
fun `test handleOnEnter`(){
   val stateSnapshot : State<MyState> = ...
   
   val expectedState : MyState = ...
   
   val actualChangeState : ChangeState<MyState> =  handleOnEnter(stateSnapshot)

  val actual : MyState = actualChangeState. ???  // How do we get the next state out of a ChangeState object ? 
  
  assertEquals( expectedState , actual)
}
```


Maybe we need to lift `reduce()` up as an `abstract fun` into `ChangeState`:


```kotlin
sealed class ChangeState<S : Any> {
   abstract fun reduce(state : S) : S
}
```

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -295,87 +224,70 @@ public class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
...
-                handler.handle(value, state)
-                NoStateChange
+            handler = { value: T, state: State<InputState> ->
+                handler(value, state.snapshot)
+                InternalNoStateChange
             })
     }
 
-    public fun <SubStateMachineState : S> stateMachine(
+    public fun <SubStateMachineState : Any> onEnterStartStateMachine(
```
sockeqwe, line 241: 👍

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Allows to create [ChangeState] objects to change the state as a result of DSL
+ * methods like [InStateBuilderBlock.on] or  [InStateBuilderBlock.onEnter].
+ *
+ * The [snapshot] property can be used to access the current state at the time the action
+ * or event happened. Note that it might be outdated by the time the code runs and should
+ * never be used within [mutate] or to create an object that is passed to [override]
+ */
+public class State<InputState : Any>(
```
sockeqwe, line 11: I can imagine that people call their State classes with `State` suffix already. i.e. `LoadingState`  or `FooState`.
Then we have `State<LoadingState>` which reads a bit strange.

Maybe we could call the class `Snapshot` instead of `State`, then it would become `Snapshot<LoadingState>` which sounds in my ears slightly better.


```kotlin
public class Snapshot<InputState : Any> (
   val state : InputState // or maybe name this property "current" instead of "state" or just "value"
) { 
  ...  
} 
```

With `Snapshot` as name then probably `.mutate()` or `.override()` don't read that nice anymore ...

Also as we have discussed on slack, maybe in a follow-up iteration we might want to get to a point where return type of "handler" is also `State`. That could read a bit strange with renaming it to `Snapshot`( but also not terrible bad ) :  `fun handleOnEnter( snapshot : Snapshot<MyState>) : Snapshot<MyState>`


What do you think?

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
```diff
@@ -0,0 +1,37 @@
...
+    val snapshot: InputState
+) {
+    /**
+     * Use this function if you want to "mutate" the current state by copying the old state
+     * and modify some properties in the copy of the new state. A common use case is to call
+     * .copy() on your state defined as data class.
+     */
+    public fun <S : Any> mutate(reducer: InputState.() -> S): ChangeState<S> {
```
sockeqwe, line 19: Do we want to introduce 2 different methods? one `mutate(reducer : InputState.() -> InputState)` and one `override(InputState.() -> S)` and a shortcut `override(newState : S)`. I think we also discussed in this context to maybe rename `override` to  `transition`. 

I have not strong opinion, just want to ask if you have considered it and explicitly would like to stick with what is part of this PR.

sockeqwe (approved):
Looks good 👍  

I have left some comments / thoughts

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
gabrielittner: Also in general

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
gabrielittner: Looks like it was already not possible to do this? We should definitely change that and I think putting reduce directly into ChangeState makes the most sense

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
gabrielittner: Having two methods was a solution for not having `InputState` anywhere but `State` already solves that. With that solved we don't need an extra method with a more concrete return type as the return type can be inferred by the compiler anyways, so you should never need to specify `S` specifically. The compiler will also guarantee that `S` is a super type of `InputState` when you use `mutate` or a method returning the result of `mutate` inside the DSL even though we don't have the `InputState : S` hierarchy encoded directly in `State`. 

So if we introduce an extra method it would simply be because we think it looks better when reading it and I don't have a strong opinion on that (same for whether we should call it override or transition).

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
gabrielittner: I took a cursory look across the freeletics code base and most of our state subclasses don't have state in the name (parent class FooState but then its subclasses are just Loading/Error/Content), but I still get your point.

Renaming it to something else is fine for me, but I'm not sure about calling it `Snapshot`. As you said the methods don't read as nicely and also if we think about replacing the `ChangeState` class I think returning a snapshot is weird. The snapshot we have is a state that was the current state when your dsl method is called but might be not the current state anymore while your method executes and especially not when the `mutate` lambda is invoked. What you return is a future change to the state, I think that would work with the `State` naming and maybe others but with `Snapshot` it's confusing.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
gabrielittner: Will do this in a follow up

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
sockeqwe: Correct, it's purely about readability (with mutate I expect a "change" or mutation of the current "state value" and not returning an entirely different state object. 

Moreover, if we go with `Snapshot` as nane I would be in favor of renaming override to transition and introduce the 2 options variations `transition(newState : S)` and `transition(reduce: InputState.() -> S)` and keep mutation only for mutating same type

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
sockeqwe: Then let's keep `State`, I also dont have a better name ...

Then I think keeping `override` as name makes sense. I would still introduce 2 overload for `override()` as discussed above and keep mutate for pure "mutation"

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
sockeqwe: 👍 Very strange, I do remember having written some tests like that a year ago or so in another side project, just try to remember which one 😃 so that I could check it ...

But I guess I never updated that side project then ...

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
gabrielittner: Changed `mutate` to return `InputState` and added an overload of `override` here [a1ba193](https://github.com/freeletics/FlowRedux/pull/306/commits/a1ba1937ad1d66410195f56894786906a33d0d5f)

sockeqwe:
I am just thinking out loud: This PR introduces a lot of breaking changes anyway (all "handlers" have now `State` as parameters), therefore maybe we can go even one step further and remove all the deprecated methods and `ChangeState` classes ...

gabrielittner:
For me it was less about migrating in multiple steps and more about making the one step migration a bit more pleasant by having less things that are suddenly not found anymore.

PR #308, opened 2022-06-04
only allow one state collector at a time

gabrielittner:
As discussed in #226, the most reliable behavior is to not allow collecting state multiple times at once. In that case only one of the flows would receive actions and it's not predictable which one would.

I also inlined the higher level `reduxStore` method into the state machine. It was already internal and only used there and I think it's easier to understand `FlowReduxStateMachine` this way without jumping around.

Closes #226


flowredux/src/commonTest/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachineTest.kt
```diff
@@ -88,4 +90,43 @@ class FlowReduxStateMachineTest {
 
         assertEquals(expectedMsg, exception.message)
     }
+
+    @Test
+    fun `observing state multiple times in parallel`() = suspendTest {
```
sockeqwe, line 95: Rename to something like  `observing state multiple times in parallel throws exception`() to make clear it should not be allowed to collect multiple times.

sockeqwe (approved):
LGTM 👍

flowredux/src/commonTest/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachineTest.kt
gabrielittner: 👍 
```suggestion
    fun `observing state multiple times in parallel throws exception`() = suspendTest {
```

PR #318, opened 2022-06-27
cleanup state

gabrielittner:
This removes the deprecated ChangeState subclasses which means the new methods on state are now the only way to create ChangeState objects. It also removes `state.override(newState)` to force using the lambda version `state.override { ... }`. The former leads to the new state often being created by reading the snapshot which can lead to bugs because the snapshot can be outdated. The version with lambda gives you the current state as `this` and makes this less likely from what I could observe. Also makes it easier to implement the lint checks because we can enforce that new state is created inside the lambda and that the snapshot isn't accessed in there.


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
```diff
@@ -1,45 +1,15 @@
...
- * - [NoStateChange] to indicate that this function did not change the state at all. There should be only very little
- * use case for [NoStateChange]. So only use it if you actually know what you do.
+ * Represents a state transition. Instances of this a created through the [State.mutate],
+ * [State.override] and [State.noChange] methods.
  *
  * [ChangeState] doesnt allow you to directly access any property.
  * Then you may wonder how do you write unit test for one of your functions that return a [ChangeState]?
  * You need to call [ChangeState.reduce] to get the actual result of the change state.
  */
 public sealed class ChangeState<out S>
```
sockeqwe, line 11: Unrelated to this PR, but I am wondering if we should name this `ChangedState`?

My reasoning is then the method signatures of handlers reads more fluent:
```fun
fun onFooAction(action : Foo, state : State<MyState<) : ChangedState { ... }
```

So you  read it "I get State and return a changed state". 

What do you think? 
Or do you think we are  moving fast to remove `ChangeState` completely and have `State<MyState>` also as return type instead of `ChangedState`? I haven't spend much time yet to think about it ...

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
```diff
@@ -48,25 +18,17 @@ internal class UnsafeMutateState<InputState, S>(
         reducer(state as InputState)
 }
 
-/**
- * No change, this is semantically equivalent to use [OverrideState] and pass in the previous state
- */
-@Deprecated("call noChange() on State instead", level = DeprecationLevel.WARNING)
-public object NoStateChange : ChangeState<Nothing>()
+internal data class OverrideState<S>(internal val newState: S) : ChangeState<S>()
```
sockeqwe, line 21: 👍 great that yoou made this class and siblings `internal`

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -139,7 +139,7 @@ public class InStateBuilderBlock<InputState : S, S : Any, A : Any>(
     ) {
         onEnter { state ->
             handler(state.snapshot)
-            InternalNoStateChange
+            NoStateChange
```
sockeqwe, line 142: No need to change anything but I am just wondering if we should call `state.noChange()` instead of `NoStateChange´ directly to only have in the future one single point that we need to change ... but could also fire back (i.e. if we do intend only to make "internal" changes), so I would probably keep it as you have it. So everything is fine ... I was just  thinking out loud ...

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
```diff
@@ -26,25 +27,17 @@ public class State<InputState : Any>(
...
      */
     public fun <S : Any> override(reducer: InputState.() -> S): ChangeState<S> {
         return UnsafeMutateState(reducer)
     }
 
     /**
-     * Sets a new state by directly override any previous state. If the new state requires values
-     * from the current state prefer using the `override` function that has a `reducer` parameter.
-     */
-    public fun <S : Any> override(newState: S): ChangeState<S> {
```
sockeqwe, line 39: I know that we spoke about removing or replacing this with `.override { ...}` but just for the records (and because I have a terrible memory): what do we want to achieve here with removing this?

sockeqwe: I'm fine with removing it, I just cant remember how our discussion about this went 😄

sockeqwe (approved):
Nice 🎉

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
gabrielittner: Also not sure about the discussion, the general motivation is what I wrote in the description
> It also removes state.override(newState) to force using the lambda version state.override { ... }. The former leads to the new state often being created by reading the snapshot which can lead to bugs because the snapshot can be outdated. The version with lambda gives you the current state as this and makes this less likely from what I could observe. Also makes it easier to implement the lint checks because we can enforce that new state is created inside the lambda and that the snapshot isn't accessed in there.

During the migration to `State` I saw others automatically replacing something like
```
val newState = MyState(stateSnapshot.something, ...)
return MutateState { newState }
```
with the correct approach of building the state inside `mutate` or `override`. The simple `override` still invites to doing the wrong approach above.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ChangeState.kt
gabrielittner: ChangedState sounds good, I can update this here.

> Or do you think we are moving fast to remove ChangeState completely and have State<MyState> also as return type instead of ChangedState? I haven't spend much time yet to think about it ...

I also haven't though to much about whether using State everywhere could have some unintended downsides or side effects

sockeqwe: Then let's go with ChangedState for now

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/State.kt
sockeqwe: Ah yea, thanks for the reminder 🙏

PR #330, opened 2022-07-06
Add DSL marker

sockeqwe:
I am not sure if I do the right thing here to be honest


gabrielittner (approved):

sockeqwe:
Let's merge this now.


One could still write 

```
inState <S>{

  onEnter { 
       onEnter {
       
       }
   }
}
```


@gabrielittner any idea what needs to be annotated to prevent this?

PR #344, opened 2022-07-31
skip publishing snapshots when secrets are not available

gabrielittner:
Will test that this actually works with one of the open dependabot prs

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -8,8 +8,6 @@ on:
 
 jobs:
   publish:
-    if: ${{ github.actor != 'dependabot[bot]' }}
-
     runs-on: macos-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-snapshot.yml
```diff
@@ -41,3 +39,4 @@ jobs:
           ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
           ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.MAVEN_SIGNING_PRIVATE_KEY }}
           ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.MAVEN_SIGNING_PASSWORD }}
+        if: "${{ env.ORG_GRADLE_PROJECT_mavenCentralUsername != '' }}"
```
sockeqwe, line 42: shouldn't that one go up to line 12 right after `publish` (so basically replace `if: ${{ github.actor != 'dependabot[bot]' }}`

gabrielittner: That doesn't work because the environment variable doesn't exist there

sockeqwe: 👍

sockeqwe (approved):

PR #352, opened 2022-08-21
add dependency analysis and binary compatibility plugins

gabrielittner:
Lets us spot breaking api changes and missing/additional dependencies


sockeqwe (approved):
Good idea 👍💡

PR #367, opened 2022-09-07
lint: replaced if/throw in favour of check{}

mcatta:
It makes the code a bit more readable, `check() {}` throws anyway `IllegalStateException` as before.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html


sockeqwe:
Thanks for your PR. Highly opinionated but I personally don't think that having `check` instead of a simple `if ... throw IllegalStateException` makes a difference or the existing code better. I would even argue that `check` requires the reader and maintainer of this code to have knowledge that `check {...}` throws an `IllegalStateException`, whereas having a simple `if ... throw` might be a bit more character compared to just `check {...}` but it `if ... throw` is very explicit and readable imho.

Therefore, I tend not to accept your suggested change.

What do you think?

mcatta:
Hi, yes probably depends on how often a specific syntax is used, I agree with you that `if/throw`is more common and maintainable. 👍

hoc081098:
I think `check` is more readable 😂

mcatta:
As you wish XD, I mean there isn't a standard for that, depends witch code style we want to use. I'm fine for the if/throw. If you agree I can close the PR

sockeqwe:
Let's keep the  `if/else`. I prefer that one.

Thanks anyway for your contribution 🙇

PR #385, opened 2022-11-04
optional values in action mapper

josirichter:
This allows for the action mapper in `onEnterStartStateMachine` to map to optional values. This way we can avoid triggering the sub state machine unnecessarily when not all actions of the main state machine are needed.


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/StartStatemachineOnEnterSideEffectBuilder.kt
```diff
@@ -52,10 +52,14 @@ internal class StartStatemachineOnEnterSideEffectBuilder<SubStateMachineState :
...
                                 coroutineScope {
                                     launch {
+                                        actionMapper(action.action)?.let {
                                         // safety net:
                                         // if sub statemachine is null then flow got canceled but
                                         // somehow this code still executes
-                                        subStateMachine?.dispatch(actionMapper(action.action))
+                                            subStateMachine?.dispatch(
+                                                it
+                                            )
```
gabrielittner, line 61: nit: 
```suggestion
                                            subStateMachine?.dispatch(it)
```

gabrielittner (commented):
Could you do the same for the action mapper in `onActionStartStateMachine`

gabrielittner (approved):
👍

sockeqwe:
> This way we can avoid triggering the sub state machine unnecessarily when not all actions of the main state machine are needed.

I don't understand the problem and why this is the solution for it, but got for it  😄

gabrielittner:
> I don't understand the problem and why this is the solution for it, but go for it 😄

If you don't want to forward all actions and the sub state machine actions are a different class you would need to introduce a `NoopAction` subclass that you then forward to the sub state machine and not act on.

sockeqwe:
but then it is only relevant for `onActionStartStateMachine`, right? 

because for `onEnterStartStateMachine` it is independent of actions or do I miss something?

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/internal/StartStatemachineOnEnterSideEffectBuilder.kt
```diff
@@ -52,10 +52,12 @@ internal class StartStatemachineOnEnterSideEffectBuilder<SubStateMachineState :
                                 // dispatch the incoming actions to the statemachine
                                 coroutineScope {
                                     launch {
+                                        actionMapper(action.action)?.let {
```
sockeqwe, line 55: I personally don't like the `?.let { ... }` syntax. for me a simple `if action != null` check is more readable. 

Just a personal opinion. not a blocker.

gabrielittner:
> but then it is only relevant for onActionStartStateMachine, right?
> because for onEnterStartStateMachine it is independent of actions or do I miss something?

No for all of them. Let's say your main state machine has this action class
```
interface MainAction {
  object A : MainAction
  object B : MainAction
  object C : MainAction
}
```
Now you have a sub state machine with
```
interface SubAction {
  object Foo : MainAction
}
```
The `Foo` action matches `A` but the sub state machine doesn't have anything for `B` and `C` but since the action mapper currently requires an action to be returned you would need something to map `B` and `C` to. This also happens also for the onEnter state machines because they will receive all actions the parent gets while they are running

josirichter:
> but then it is only relevant for onActionStartStateMachine, right?
> because for onEnterStartStateMachine it is independent of actions or do I miss something?

Starting the state machine is independent from actions, yes, but it can still react on them through the mapper. But the sub state machine won't handle all the triggered actions in most use cases we were approaching.

Also I discovered a rare crash that happened when an action was fired wile the sub state machine was not fully operational yet. While this change is no fix for that it further reduces the probability of it happening.

sockeqwe:
Thanks for the clarification, I understand better now and get why we need  a way to "ignore" an action. Maybe that could even be the default `actionMapper = { null }` so we can reduce a few overload functions?


Another topic: @josirichter, I have noticed that you have release version 1.0.1 .  Was this intentionally? I assume this PR should have been part of 1.0.1 too but has not been merged yet ...

sockeqwe:
No worries, could we also please update the Docs (plus add unit tests)

josirichter:
So I added some tests and adapted the Docs 👌

flowredux/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StartStateMachineOnActionInStateTest.kt
```diff
@@ -303,6 +303,75 @@ class StartStateMachineOnActionInStateTest {
         }
     }
 
+    @Test
+    fun `actions are only dispatched to sub statemachine if they are mapped`() = suspendTest {
```
sockeqwe, line 307: nitpicking: maybe rename it something a bit more expressive. maybe something like 

`actions are only dispatched to sub statemachine if they are not mapped to null`()  (I think this "null" part would make it a bit more explicit for me

flowredux/src/commonTest/kotlin/com/freeletics/flowredux/dsl/StartStateMachineOnActionInStateTest.kt
```diff
@@ -303,6 +303,75 @@ class StartStateMachineOnActionInStateTest {
...
+            assertEquals(parentActionInvocations, 3)
+
+            // dispatch unmapped A3 action
+            parent.dispatch(TestAction.A3)
+            delay(10) // give child a bit of time before continuing
+            // assert that child state machine was not triggered after first two initial actions
+            assertEquals(childActionInvocations, 1)
+            assertEquals(parentActionInvocations, 4)
+        }
+    }
```
sockeqwe, line 374: Great test case 💪

sockeqwe:
@josirichter  left some comments, but looks good! 

in general my "approval" is not needed in this repository. Just go ahead and merge / change things if you feel it is the right thing to do. If you want to consult / get input from me, feel free to ping me, but for future reference, don't see me as a blocker or someone you need to get approval from.

gabrielittner (approved):

PR #397, opened 2022-12-10
update tests and make latest turbine + runTest work

gabrielittner:
- update turbine to 0.12.1 (closes #355)
- update Kotlin to 1.7.20 and Compose Compiler 1.3.2
- replace `suspendTest` with `runTest`
- remove delays from tests and use channels for synchronization


gabrielittner:
Downgraded Gradle because of https://youtrack.jetbrains.com/issue/KT-54634

gabrielittner:
runTest isn't strictly required I mostly went with it because it is what coroutines provides for tests that run coroutines

sockeqwe (approved):

PR #398, opened 2022-12-10
make methods with KClass parameter internal with @PublishedApi

gabrielittner:
With this these methods that only exist to be called from the reified inline methods are hidden from auto completion since they are internal but still accessible to the inline functions. This is a binary compatible change (see `flowredux.api` file, the change there comes from removing the `@JvmSynthetic`)


sockeqwe (approved):
👍 nice

PR #399, opened 2022-12-10
allow any StateMachine to be a sub state machine

gabrielittner:
My idea is to migrate some bigger older state machines in our code base to flow redux gradually by making them a sub state machine of the flow redux replacement. To make that possible all sub state machine dsl methods now get `StateMachine` instead of `FlowReduxStateMachine`. Since it is a super interface the change should be binary compatible.


sockeqwe (approved):
Cool 😎

PR #400, opened 2022-12-10
deploy website directly from github actions

gabrielittner:
This makes the gh pages branch obsolete. Instead of building the site and pushing it to that branch the action uploads it and triggers a deployment. We've been using it in mad for a while now https://github.com/freeletics/mad/blob/main/.github/workflows/publish-docs.yml

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -10,11 +10,20 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    permissions:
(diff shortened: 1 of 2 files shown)


sockeqwe (approved):
I did not know that such dirct deployments without gh pages branch work

gabrielittner:
It's technically still in beta but I haven't seen any issues with it yet

Screenshot 2022-12-11 at 16 46 32

PR #405, opened 2023-01-02
automate more parts of the publishing process

gabrielittner:
With this it is not neccessary to push commits to `main` anymore to do a release and as much as possible is automated. The remaining steps are:
1. update changelog (through a pr), we could probably also have some automation here but I personally don't like these auto generated changelogs
2. create and push a tag for the new version

To make this possible there are a few changes:
- we now have a `CHANGELOG.md` file comitted to the repo
- the release workflow will extract the section for the new release and automatically create a Github release with it
- `VERSION_NAME` is not comitted anymore and instead always computed
- for releases we tag the tag name as version
- for snapshots we take the last tag increment the last number by 1 and then append SNAPSHOT
- the readme contains `<latest-version>` pacehoders and one of those badges that shows the last version
- for the website we automatically replace `<latest-version>` with the last tag and remove the badge
- snapshot and website workflows run after the release so that there is a snapshot for the next release and the website contains the latest version

The only part that I don't like that much is the `<latest-version>`  placeholder in the README, I prefer when it just has the full declaration ready to copy, but at least it's like this on the website.

Already tested this [here](http://github.com/gabrielittner/renderer)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,10 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
 
 jobs:
   build:
@@ -30,4 +34,4 @@ jobs:
       - uses: gradle/gradle-build-action@v2
 
       - name: Build with Gradle
-        run: ./gradlew build --stacktrace
+        run: ./gradlew build buildHealth --stacktrace
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -4,11 +4,17 @@ on:
   push:
     branches:
       - main
+  # TODO: replace with on push tag https://github.com/actions/deploy-pages/issues/76
+  workflow_run:
+    workflows: [Publish Release]
+    types:
+      - completed
 
 jobs:
   publish:
 
     runs-on: ubuntu-latest
+    if: github.repository == 'freeletics/flowredux'
 
     permissions:
       contents: read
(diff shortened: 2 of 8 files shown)


sockeqwe:
👋 great idea!
I am thinking even one step forward: have you considered using github actions with input fields? With that we could even simplify `CHANGELOG.md` and pushing tags.

My idea is:

Instead of changing `CHANGELOG.md` manually and creating a PR, we could create a "release" github action that can be started manually from github action section on github.com

And this action has 2 text input fields: 
- a version number input field. This is used for setting the version for release and as title fir changelog entry.
- a text field to describe the changes. This is used as description to add on top of CHANGELOG.md file along eith the version number from first text input field.

The rest remains as described by your proposal more or less but is auto generated by the github action: automatically changes changelog file and creates a tag (and even a release on guthub through rest api). Also we don't need to create a PR for reviewing changelog and releasing it (unless we really want to have this as a manual step which I don't like to have).


What do you think?

gabrielittner:
The issue with that is that is that is that Github Actions can't push to a protected branch. I initially had something like this that would set the `UNRELEASED` in the changelog to today's date. 

I think the changelog is fine as long as we keep it up to date while making changes for the upcoming release (which is a good idea anyways since we have very infrequent releases, so you don't need to remember what that change from 2 month ago meant when trying to write a changelog). Then the changelog update when releasing is just setting the date and that could also be done in parallel to the release so that you don't have to wait for someone to approve

gabrielittner:
Merged this as is for now because it's working well in other repos. We can still automate more on top of this

PR #445, opened 2023-03-23
fix flaky tests

gabrielittner:
These were flaky because the state emission might happen before the new state is fully propagated internally, so while we receive `assertEquals(TestState.S2, awaitItem())` the cancellation of the old state might not have happened yet. The `signal.close()` might then also happen before that which causes `reached` to become `true`. The delay isn't nice but it causes the internal coroutines before which then causes the block to be cancelled during the delay. (note that this is not a real time delay, just virtual time in the test scheduler, so the time value there doesn't matter). I've tested both of them by running them 1k times each.


fraherm (approved):
LGTM

PR #447, opened 2023-03-23
update mkdocs config

gabrielittner:
Applies the same config changes that I did before in mad (see https://github.com/freeletics/mad/pull/278) and adds the chagelog to the website.


fraherm (approved):
LGTM

PR #448, opened 2023-03-23
use freeletics gradle plugins

gabrielittner:
Reduces the amount of manual config that is needed. I had to resolve some warnings in the sample project as part of this because we default to fail on warnings.


fraherm (approved):
LGTM

PR #456, opened 2023-03-27
minor iOS sample improvements

jalalawqati:
Just a few small changes here and there to the iOS sample project, mainly adding more previews and slightly enhancing syntax.

Also added an error view and used it whenever an error state is emitted.

### DIFF
--- a/sample/ios/ios/ContentView.swift
+++ b/sample/ios/ios/ContentView.swift
@@ -18,23 +18,25 @@ struct ContentView: View {
     )
 
     var body: some View {
-        
-        return ZStack {
-            if state is LoadFirstPagePaginationState {
-                LoadingIndicatorView()
-            } else if let state = state as? ShowContentPaginationState {
-                GithubReposList(
-                    contentState: state,
-                    dispatchAction: dispatchAction
-                )
-            } else if state is LoadingFirstPageError {
-                // TODO extract standalone widget?
-                Button(action: triggerReloadFirstPage) {
-                    Text("An error has occured.\nClick here to retry.")
-                }
-            }
-        }.onAppear(perform: startStateMachine)
+        contentView()
+            .onAppear(perform: startStateMachine)
 
+    }
+
+    @ViewBuilder
+    private func contentView() -> some View {
+        switch state {
+        case is LoadFirstPagePaginationState:
+            LoadingIndicatorView()
+        case let state as ShowContentPaginationState:
+            GithubReposList(contentState: state, dispatchAction: dispatchAction)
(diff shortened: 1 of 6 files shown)


sample/ios/ios/ContentView.swift
```diff
@@ -18,36 +18,36 @@ struct ContentView: View {
...
-                }
-            }
-        }.onAppear(perform: startStateMachine)
+        contentView()
+            .onAppear(perform: startStateMachine)
 
     }
 
+    @ViewBuilder
+    private func contentView() -> some View {
```
sockeqwe, line 27: Nice 👍

sample/ios/ios/ContentView.swift
```diff
@@ -18,36 +18,36 @@ struct ContentView: View {
...
+        switch state {
+        case is LoadFirstPagePaginationState:
+            LoadingIndicatorView()
+        case let state as ShowContentPaginationState:
+            GithubReposList(contentState: state, dispatchAction: dispatchAction)
+        case is LoadingFirstPageError:
+            ErrorView(action: triggerReloadFirstPage)
+        default:
+            EmptyView()
```
sockeqwe, line 36: Suggestion: I would actually raise an Error here. That would help catch "unhandled" cases in the future instead of "wondering" why things are not displayed. 

But it is your call ...

sample/ios/ios/ContentView.swift
```diff
@@ -18,36 +18,36 @@ struct ContentView: View {
...
+
     private func triggerReloadFirstPage() {
-        self.stateMachine.dispatch(action: RetryLoadingFirstPage())
+        stateMachine.dispatch(action: RetryLoadingFirstPage())
     }
     
     
     private func dispatchAction(action : Action){
-        self.stateMachine.dispatch(action: action)
+        stateMachine.dispatch(action: action)
```
sockeqwe, line 46: uh, I didnt know that this works 🎉  
I thought an explicit `self` is needed here to access `stateMachine` 👍

sample/ios/ios/ErrorView.swift
sockeqwe, line 1: Nice 💪

sockeqwe (approved):
Looks good. Thanks 👍

sample/ios/ios/ContentView.swift
jalalawqati: Nice suggestion 👍  -> https://github.com/freeletics/FlowRedux/pull/456/commits/130651fbd07e8e6aace0c5b502ce4be0156f6476

PR #478, opened 2023-04-28
switch ktlint to kts script

gabrielittner:
To keep things aligned this is now using our ktlint script instead of the Gradle plugin.


kboyarshinov (approved):
:+1:

Issue #467, opened 2023-04-18
Cannot dispatch action because state Flow of this FlowReduxStateMachine is not collected yet


sarahborgi:
Hi, w're using FLowRedux state machine in our code and we have been getting non systematic failures always happen in the same transition point, Here's a simple exemple:

```kotlin
internal sealed class State {

    object Idle : State()
    object First : State()
    object Second : State()
    data class Third(
        var subState: State(),
        val parentStateMachine: FlowReduxStateMachine<State, Event>? = null
    )
    object One : State()
    object Two : State()
    object Three : State()
}

sealed class Event {
    object ToFirst : Event()
    object ToSecond : Event()
    object ToThird : Event()

    object ToTwo : Event()
    object ToThree : Event()

    object GoBack : Event()
}

@OptIn(FlowPreview::class, ExperimentalCoroutinesApi::class)
internal class FirstStateMachine(
) : FlowReduxStateMachine<State, Event>(initialState = State.First)

val currentFlow: FlowReduxStateMachine<State, Event> = this
{

    init {
        spec {
            inState<State.Idle> {
                on {
                    _, Event.ToFirst ->
                    state.override { State.First }
                }
            }

            inState<State.First> {
                on {
                    _, Event.ToSecond ->
                    state.override { State.Second }
                }
            }

            inState<State.Second> {
                on {
                    _, Event.ToThird ->
                    state.override { State.Third(One, currentFlow) }
                }
            }
            inState<State.Third> {
                onEnterStartStateMachine(
                    stateMachineFactory = {
                        SecondStateMachine()
                    },
                    stateMapper = { state, subState ->
                        state.override { subState }
                    },
                    actionMapper = {
                        it
                    }
                )
            }
            inState<State> {
                on { _: Event.GoBack, state ->
                    state.override { State.Idle }
                }
            }
        }
    }
}

@OptIn(FlowPreview::class, ExperimentalCoroutinesApi::class)
internal class SecondStateMachine(
) : FlowReduxStateMachine<State, Event>(initialState)

val currentFlow: FlowReduxStateMachine<State, Event> = this
{

    init {
        spec {
            inStateWithCondition({ state: State.Third ->
                state.subState == State.One
            })
            {
                on { _: Event.toTwo, state ->
                    state.mutate { copy(subState = State.Two) }
                }
            }

            inStateWithCondition({ state: State.Third ->
                state.subState == State.Two
            })
            {
                on { _: Event.toThree, state ->
[...]

PR #482, opened 2023-05-02
Fixes issues where Actions are dispatched to substatemachine which is not collected yet

sockeqwe:
Sometimes we dispatch Actions to sub statmachines before the state (Flow) of the substate machine is collected. This leads to errors like #467 

The underlying issue is that we don't synchronize coroutines for dispatching actions and collecting the `Flow<State>`.

This PR introduces a simple `CoroutineWaiter` class that helps to synchronize coroutines. We use it to suspend any Coroutine that wants to dispatch an `Action` to the sub statemachine until the sub statemachine flow is actually collected. So dispatching actions is only resumed and done once the `Flow<State>` is guaranteed to be collected.

For now I have implemented it only for the `onEnterStartStateMachine { ... }` DSL part but if we think this is the right direction to go, then I could move ahead and reuse `CoroutineWaiter` also for `onActionStartStatemachine { ... }` .


gabrielittner (approved):
Thanks 👍

PR #487, opened 2023-05-06
Fix flaky on Test: actionBlockStopsWhenMovedToAnotherState

sockeqwe:
The problem with the previous implementation was that it can take a few ms for FlowRedux to cancel the `on<Action> {...}` block. therefore we have introduced the  `delay(100)` but since we did `advanceUntilIdle()` and `runCurrent()` on the `TestDispatcher` it skips the delay and it was sometimes still too fast (so cancelation happened afterward). That resulted in a flaky test.

Now I rephrased the test towards "Block is canceled within 10 milliseconds after having moved to another state"


gabrielittner (approved):
LGTM, I think there are some more tests like this

PR #496, opened 2023-05-30
some cleanup

gabrielittner:
While preparing and refactoring some things before adding the condition and identity blocks I've noticed some things. 

- moved `ChangedState` to `State` since they belong together
- moved `AtomicCounter`, `CoroutineWaiter` and flow methods to `util`
- all side effect builders have been moved from `internal` to a `sideeffects` package (also `Action` since it's related to them but that will be removed anyways during the refactoring)
- removed some obsolete comments/docs/todos


sockeqwe:
I am on vacation right now and cannot do 2FA to github at the moment (I
reply via email).

Overall looks good (from what I can see from my mobile phone) 👍 but CI has
failed. You may wanna check the logs

Gabriel Ittner ***@***.***> schrieb am Mi., 31. Mai 2023,
03:53:

gabrielittner:
There was one thing thing caught by the binary compatibility validator, which is resolved now.

Have a nice vacation

PR #501, opened 2023-06-12
make reduxStore more state machine specific

gabrielittner:
The current `reduxStore` implementation is still like it was when it was the standalone core of the library. This is a first step away from that and towards transforming it into something specifically built for `FlowReduxStateMachine`. 

We were wrapping `A` into `Action<S, A>` inside the state machine and the `A` of the `reduxStore` was then `Action<S, A>`. This then also applied to the state machine. With this `reduxStore`, `SideEffect` and related types will directly use `Action` meaning that `A` will be the same everywhere and not sometimes the wrapped type.

Other changes:
- move the action wrapping to the store
- after a state change broadcast `InitialStateAction` instead of the `ChangedStateAction` that caused the state change (the latter will be completely removed in a follow up)
- inline the reducer since we always have the same one
- removed the `reduxStore` specific test, we're covering this part through all the other tests and the value of these tests will further decrease in follow ups when implementation and behavior of `reduxStore` will change a lot


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
```diff
@@ -40,12 +34,7 @@ public abstract class FlowReduxStateMachine<S : Any, A : Any>(
 
         outputState = inputActions
             .receiveAsFlow()
-            .map<A, Action<S, A>> { ExternalWrappedAction(it) }
-            .onStart {
-                emit(InitialStateAction())
-            }
-            .reduxStore(initialStateSupplier, sideEffects, ::reducer)
-            .distinctUntilChanged { old, new -> old === new } // distinct until not the same object reference.
```
sockeqwe, line 48: dont we need the distinct until changed?

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/Action.kt
```diff
@@ -6,9 +6,16 @@ import com.freeletics.flowredux.dsl.reduce
 internal sealed class Action<S, A>
```
sockeqwe, line 6: nitpicking: but we could make this non null:`S:Any, A:Any`

sockeqwe (approved):
looks good 👍

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStateMachine.kt
gabrielittner: We have this check inside the store now https://github.com/freeletics/FlowRedux/pull/501/files#diff-0bd32fee6f00b61666521e8846b321dce06c0dbf4ea5266f85bc6a7abd741e6bR42 It has the advantage that we also don't run the side effects again if the state stayed identical.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/Action.kt
gabrielittner: I'll get rid of this completely, so I kept it like this for now. It would trigger a ton of changes for SideEffect and the builders as well

PR #502, opened 2023-06-12
introduce BaseBuilderBlock

gabrielittner:
Since we'll add `condition` and the identity block that share all the existing DSL methods this introduces a base class that can be used for all 3 block types that we'll have. According to [this](https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_Interfaces) it is a binary compatible change: `Move API method up type hierarchy -> If method in supertype need not be implemented by Client -> Binary compatible`. 

The base class interally exposes the in state builders instead of side effects in preparation of refactoring these parts more to go away from our initial redux implementation.


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStoreBuilder.kt
```diff
@@ -1,20 +1,14 @@
...
 
-import com.freeletics.flowredux.SideEffect
+import com.freeletics.flowredux.sideeffects.InStateSideEffectBuilder
 import kotlin.reflect.KClass
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 
 @ExperimentalCoroutinesApi
 @FlowReduxDsl
-public class FlowReduxStoreBuilder<S : Any, A : Any> {
+public class FlowReduxStoreBuilder<S : Any, A : Any> internal constructor() {
```
gabrielittner, line 9: This is technically a breaking change but there isn't really a point in anyone using the constructor since they can't do anything with the created instance.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/InStateBuilderBlock.kt
```diff
@@ -1,327 +1,9 @@
...
-        inStateSideEffectBuilders += StartStateMachineOnActionInStateSideEffectBuilder(
-            isInState = isInState,
-            subStateMachineFactory = stateMachineFactory,
-            subActionClass = actionClass,
-            actionMapper = actionMapper,
-            stateMapper = stateMapper,
-        )
-    }
-}
+public class InStateBuilderBlock<InputState : S, S : Any, A : Any> internal constructor(
```
gabrielittner, line 7: Same here that the constructor is now marked as internal

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStoreBuilder.kt
```diff
@@ -1,20 +1,14 @@
...
 
-    private val builderBlocks: MutableList<InStateBuilderBlock<*, S, A>> = ArrayList()
-
-    internal fun generateSideEffects(): List<SideEffect<S, A>> {
-        return builderBlocks.flatMap { builder ->
-            builder.generateSideEffects()
-        }
-    }
+    internal val sideEffectBuilders: MutableList<InStateSideEffectBuilder<out S, S, A>> = ArrayList()
```
sockeqwe, line 11: Nitpicking: Should this really be mutable List from a public API (although internal modifier) point of view

sockeqwe (approved):

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/FlowReduxStoreBuilder.kt
gabrielittner: Was a copy paste mistake from `BaseBuilderBlock`. There it needs to be exposed because subclasses need to be able to add something and `protected` gets flag for exposing an `internal` type.

PR #504, opened 2023-06-14
add reusable stateChange to InStateSideEffectBuilder

gabrielittner:
Similar to `runOnlyIfInInputState` we now have a reusable method to produce a flow that emits a state change in the builder. This replaces the various versions of `setStateFlow` in the subclasses.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/InStateSideEffectBuilder.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/InStateSideEffectBuilder.kt
@@ -19,7 +19,7 @@ internal abstract class InStateSideEffectBuilder<InputState : S, S, A> {
 
     abstract fun generateSideEffect(): SideEffect<S, A>
 
-    protected inline fun stateChange(
+    protected inline fun changeState(
         crossinline getState: GetState<S>,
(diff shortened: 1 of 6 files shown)


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/InStateSideEffectBuilder.kt
```diff
@@ -16,6 +19,18 @@ internal abstract class InStateSideEffectBuilder<InputState : S, S, A> {
 
     abstract fun generateSideEffect(): SideEffect<S, A>
 
+    protected inline fun stateChange(
```
sockeqwe, line 22: I personally would find the wording `changeState()` a bit more readable than `stateChange()` but your call, it doesn't make code-wise any difference.

sockeqwe (approved):

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/InStateSideEffectBuilder.kt
gabrielittner: Updated [670fb26](https://github.com/freeletics/FlowRedux/pull/504/commits/670fb2689c2d12eb12c46d33a6c1fa08fe4eb578)

PR #506, opened 2023-06-15
replace SideEffect and InStateSideEffectBuilder

gabrielittner:
Until now we had `InStateSideEffectBuilder` which had a method to create return a new `SideEffect` which is in turn a transformer for a given `Flow`. All side effects produce cold flows so they could just directly return. This is the main change here. `InStateSideEffectBuilder` now just has a `produceState(actions: Flow<Action<S, A>>, getState: GetState<S>): Flow<ChangeStateAction<S, A>>` method and the `SideEffect` typealias is gone. Since the their role mostly fits and it is a simpler name `InStateSideEffectBuilder` has been renamed to `SideEffect` (happy for other suggestions).

The side effect implementatios stayed the same, the changes in them are just indentation changes.

Part 2 of the changes here that there is a new builder interface. It's called `SideEffectBuilder` and is used one level higher in the block builders. For now it doesn't do much but it comes with it's own `IsInState` interface. This second `IsInState` (SideEffect still hase the previously introduced one) is for the separation mentioned in #503. `SideEffectBuilder.IsInState` is what we have in the block builders and can then be used to check if we entered the state and which would mean that we should build the side effect and start it (to not blow it up further in this pr we're still building them once and the start/stop is still handled inside each side effect). Right now the conversion from `SideEffectBuilder.IsInState` to `SideEffect.IsInState` (to check whether we are still in the indended state) is done in `sideEffectIsInState`. Right now it is just doing the same check but the indentity block can override it and add the identity check to it (the indentity builder block implementation will basically do nothing else the rest will stay completely unchanged).

Sorry that this got bigger concept wise, the introduction of `SideEffectBuilder` probably should have been separate from this but it got a bit hard to untangle.


sockeqwe (approved):
Awesome 👍

PR #507, opened 2023-06-16
remove ChangeStateAction

gabrielittner:
With this `SideEffect` will directly emit `ChangedState` instead of the action. This completely separates the input and output of side effects and makes handling them easier. `Action` also doesn't need a state type paremeter anymore. So the flow is now `A -> Action<A> -> SideEffect -> ChangedState<S> -> Reducer -> S` with the extra that the reducer sends `InitialStateAction` whenever there is a new state.

There is still wrapping in another class going on internally (see `guardWithIsInState`), if the `ChangedState` object is not a `NoStateChange`, to continue having the reduce guarded by an `isInState` check. So this is only slightly more efficient allocation wise. I think it should be possible later to directly have this inside the originally created `ChangedState`, but it will require fighting with generics a bit sine `State` which creates them only has a type parameter for `InputState` but not one for `State`.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/FlowRedux.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/FlowRedux.kt
@@ -1,62 +1,66 @@
 package com.freeletics.flowredux
 
+import com.freeletics.flowredux.dsl.reduce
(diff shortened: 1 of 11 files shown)


sockeqwe (approved):
NICE 🚀

PR #508, opened 2023-06-16
explicitly cancel side effects before state change

gabrielittner:
This is the first bigger change that thre previous refactorings prepared for. We now have a `ManagedSideEffect` that we create in the `reduxStore` for each builder. `ManagedSideEffect` can receive `sendStateChange` when a new state is there `sendAction` for actions coming from the outside and it has an explicit `cancelIfNeeded`. Internally it will use the `SideEffectBuilder.IsInState` in `sendStateChange` to start the side effect if it wasn't started yet. The `cancelIfNeeded` method will use `SideEffect.IsInState` to cancel a running side effect if we left the state (this is where the differentiation between isInState to start and to check whether still active that I mentioned before comes in). `sendAction` will also only forward actions based on `SideEffect.IsInState` and if the side effect is already running.

The `ManagedSideEffect` with it's explicit starting and `stopping` enables various improvements:
- We explicitly call `cancelIfNeeded` before we emit the new state to consumers and before we send the state change to side effects, which fixes various issues:
    - @kboyarshinov had an issue last year where a `collectWhileInState` for the old state was only cancelled after `collectWhileInState` for the new state was already started and in this case both collected from the same object and that object only expected one collector at a time.
    - Our flaky tests are all tests for cancellation. With this when you wait for the state change with `awaitItem` you can directly afterwards assert that the block of the old state got cancelled. Iwill update the tests in a separate PR because I want to show that this change still works with the current tests.
    - The state machine behavior should become a lot more predictable. Currently the internal side effect management and how they start and stop is kind of random.
- This also enables identity changes to work. When `cancelIfNeeded` is called for a state with a different identity `SideEffect.IsInState` will become false so the running side effect is cancelled. Then `sendStateChange` is called and since the `ManagedSideEffect` doesn't have an active side effect anymore it will start a new one.

An internal change related to that is that `SideEffect` now has `sendState` and `sendAction` to receive the state and the actions from the `ManagedSideEffect`. There is temporarily a new intermediate `LegacySideEffect` which implements those 2 methods and turns them into the `Flow<Action<A>>` that our side effects expect. After this PR each side effect will one by one be re-implemented to get rid of `LegacySideEffect`, `Action<A>` and the internal start/stop logic that they have built in.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
@@ -4,9 +4,34 @@ import com.freeletics.flowredux.dsl.ChangedState
 import com.freeletics.flowredux.dsl.NoStateChange
 import com.freeletics.flowredux.dsl.UnsafeMutateState
 import com.freeletics.flowredux.dsl.reduce
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
-import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.launch
+
+internal abstract class LegacySideEffect<InputState : S, S, A> : SideEffect<InputState, S, A>() {
+
+    private val actions = Channel<Action<A>>()
+
+    override fun produceState(getState: GetState<S>): Flow<ChangedState<S>> {
+        return produceState(actions.receiveAsFlow(), getState)
+    }
+
+    abstract fun produceState(actions: Flow<Action<A>>, getState: GetState<S>): Flow<ChangedState<S>>
+
+    override suspend fun sendState(state: S) {
+        actions.send(InitialStateAction())
+    }
+
+    override suspend fun sendAction(action: A) {
+        actions.send(ExternalWrappedAction(action))
+    }
+}
 
 internal abstract class SideEffect<InputState : S, S, A> {
     fun interface IsInState<S> {
@@ -15,7 +40,11 @@ internal abstract class SideEffect<InputState : S, S, A> {
 
     abstract val isInState: IsInState<S>
 
-    abstract fun produceState(actions: Flow<Action<A>>, getState: GetState<S>): Flow<ChangedState<S>>
+    abstract fun produceState(getState: GetState<S>): Flow<ChangedState<S>>
+
+    open suspend fun sendState(state: S) {}
+
+    open suspend fun sendAction(action: A) {}
 
     protected inline fun changeState(
         crossinline getState: GetState<S>,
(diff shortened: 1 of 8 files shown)


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
```diff
@@ -15,7 +40,11 @@ internal abstract class SideEffect<InputState : S, S, A> {
 
     abstract val isInState: IsInState<S>
 
-    abstract fun produceState(actions: Flow<Action<A>>, getState: GetState<S>): Flow<ChangedState<S>>
+    abstract fun produceState(getState: GetState<S>): Flow<ChangedState<S>>
+
+    open suspend fun sendState(state: S) {}
+
+    open suspend fun sendAction(action: A) {}
```
gabrielittner, line 47: One thing I don't completely like is that it would allow a side effect implementation to call these. It's only internal so it's not that big of a problem and I couldn't come up with something better.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
```diff
@@ -63,15 +92,62 @@ internal class SideEffectBuilder<InputState : S, S, A>(
...
+
+            current.sideEffect.sendState(state)
+        }
+    }
+
+    suspend fun cancelIfNeeded(state: S) {
+        val current = current
+        if (current != null) {
+            if (!current.sideEffect.isInState.check(state)) {
+                current.job.cancel(StateChangeCancellationException())
```
gabrielittner, line 127: Having this exception here for the cancellation will mainly allow us to use it in tests to see if a block was properly cancelled.

sockeqwe: Not 100% sure if I understand how you will get the exception to be consumed inside the tests, but I guess I will see then in one of the next PRs.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
```diff
@@ -63,15 +92,62 @@ internal class SideEffectBuilder<InputState : S, S, A>(
...
+    }
+
+    suspend fun sendAction(action: A, state: S) {
+        val current = current
+        if (current != null && current.sideEffect.isInState.check(state)) {
+            current.sideEffect.sendAction(action)
+        }
+    }
+
+    private inner class Current(
```
sockeqwe, line 141: Maybe the naming could be a bit more specific. Is it the "Current Sideeffects for current State"? "Current Sideffect only"?  


Maybe you could think of a bit more expressive name to make the code a bit more readable. Up to you.

sockeqwe (approved):
Cool 😎 .

I really like the direction we are heading to.

One thing that I am not 100% sure about though:
If we always cancel the "current", could there be the case that an action is "swallowed" and not handled because the next "current" did not start yet?

Example:
1. Current --> active and running
2. State change --> Current is canceled.
3. Action dispatched (but no current running right now)
4. New Current from state transition

Could it happen that in that case the Action in step 3 got lost?

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
gabrielittner: When we test the cancellation of a block we can now do

gabrielittner: When we test cancellation of a block we can now do 
```
var caughtThrowable: Throwable? = null

//...
on<Foo> {
  try {
    awaitCancellation()
    assertFail()
  }  catch (t: Throwable) {
  }
}

// ...
assertEquals(awaitItem(), TheStateThatCancelsTheBlockAbove)
assertEquals(caughtThrowable, StateChangeCancellationException())
```
So we're not just checking that it was cancelled but we're checking that it was cancelled by us.

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
gabrielittner: What do you think of `CurrentlyActiveSideEffect`?

gabrielittner:
>  Could it happen that in that case the Action in step 3 got lost?

There is a chance for that. I'll add a mutex to redux store so that the 2 `collect` lambdas don't run in parallel

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/SideEffect.kt
gabrielittner: [1e386bd](https://github.com/freeletics/FlowRedux/pull/508/commits/1e386bd2fc94b4a22501b7f4e27890f8fe6bb085)

PR #509, opened 2023-06-16
update tests

gabrielittner:
As mentioned in #508 the tests now don't need the delay/signa/reached anymore and we can directly check that cancellation happened after we get the new state.


sockeqwe (approved):
Ah nice 💡 

I did not know about these capabilities of turbine

PR #514, opened 2023-06-17
refactor OnEnterStartStateMachine side effect

gabrielittner:
This uses the `initialState` parameter from the builder now and `ActionBasedSideEffect` to get the actions that it needs to forward.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/OnEnterStartStateMachine.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/OnEnterStartStateMachine.kt
@@ -3,99 +3,48 @@ package com.freeletics.flowredux.sideeffects
 import com.freeletics.flowredux.dsl.ChangedState
 import com.freeletics.flowredux.dsl.State
 import com.freeletics.flowredux.util.CoroutineWaiter
-import com.freeletics.flowredux.util.mapToIsInState
-import com.freeletics.flowredux.util.whileInState
 import com.freeletics.mad.statemachine.StateMachine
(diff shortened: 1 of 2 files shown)


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/OnEnterStartStateMachine.kt
```diff
@@ -3,99 +3,48 @@ package com.freeletics.flowredux.sideeffects
...
-import kotlinx.coroutines.flow.onCompletion
-import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.launch
 
 @ExperimentalCoroutinesApi
 internal class OnEnterStartStateMachine<SubStateMachineState : Any, SubStateMachineAction : Any, InputState : S, S : Any, A>(
     override val isInState: IsInState<S>,
-    private val subStateMachineFactory: (InputState) -> StateMachine<SubStateMachineState, SubStateMachineAction>,
```
sockeqwe, line 23: No factory needed anymore in the future or is that something that will come later?

sockeqwe (approved):

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/sideeffects/OnEnterStartStateMachine.kt
gabrielittner: I wanted to keep it consistent with `onEnter` where we use the initialState from the builder. We could pass both the factory and the initialState here but there isn't really a point in that.

Issue #480, opened 2023-04-30
DSL: Add a way to work with"Identity" of states


sockeqwe:
Following up on #479 it would be convenient to express staying in a certain state as long as the "identity" of a state did not change  (in addition to what we already have as DSL; this is not about removing existing DSL functionality)

Let's discuss here some ideas how we could make that happening:

## Proposal 1: Keep existing dsl and add a `withIdentity { ... }` block:

```kotlin
inState<State>  {
    on<Action> {...} // stay as it is
    onEnter {...} // stay as it is
    collectWhileInState {...} // stay as it is
    onEnterStartStateMachine { ... } // stay as it is
    onActionStartStateMachine { ... } // stay as it is

    withIdentity( identity = { state -> state.id }) {  // new DSL Block. It keeps track of whenever state's identity change
            onEnter {...} // same functionality as usual
            collectWhileInState {...} // same functionality as usual
            onEnterStartStateMachine { ... } // stay as it is

            // I think on<Action> and onActionStartStateMachine should not be part of the withIdentity block
     }
}
```
Something like the one above  👆  should be doable without breaking changes and stay backward compatible. 
Some more thoughts: 
- `withIdentity`: Probably there are better names, something like `onIdentityChanged` or something else.
- I am not sure if `Action` related DSL constructs should be part of the `withIdentity` block. I am not sure if withIdentity and Action  work well together. 
 

We could also think a bit broader and do some (breaking changes) with `inState( additionalCondition = { ... })`: Maybe something like this:

```kotlin
inState<State> { // no additionalCondition as parameter here
    on<Action> {...} // stay as it is
    onEnter {...} // stay as it is
    collectWhileInState {...} // stay as it is
    onEnterStartStateMachine { ... } // stay as it is
    onActionStartStateMachine { ... } // stay as it is

    // NEW
    condition( { state ->  .... }) {
        on<Action> {...} // stay as it is
        onEnter {...} // stay as it is
        collectWhileInState {...} // stay as it is
        onEnterStartStateMachine { ... } // stay as it is
        onActionStartStateMachine { ... } // stay as it is

        // should we allow nested conditions?
        condition( { state ->  .... }) {
           on<Action> {...} // stay as it is
           onEnter {...} // stay as it is
           collectWhileInState {...} // stay as it is
           onEnterStartStateMachine { ... } // stay as it is
           onActionStartStateMachine { ... } // stay as it is

           // Nested condition
           condition( { state -> ... } ) {
              ...
[...]

PR #517, opened 2023-06-17
implement condition block, deprecate obsolete inState methods

### LINKED ISSUE #310: DSL for additional conditions
Just an idea that I have mixed feelings about, so dropping my idea here to hear opinions.

```kotlin
enum class MyEnum { A, B, C }
data class MyState(val e : MyEnum)


inState<MyState>(additionalCondition = { it.e == MyEnum.B }) { // means state is MyState AND state.e == B
    ... 
}
```

I am wondering if this could be expressed somehow with DSL constructs. 

For example something like adding a new `condition` DSL block like this:

```kotlin
inState<MyState> {

    condition( {it.e == MyEnum.B} ) {
        onEnter {...}
        on<Action> {...}
        collectWhileInstate { ... }

        condition (...){
              // even subcondition could be supported
         }
     }
   
    
    onEnter {...}
    on<Action> {...}
    collectWhileInstate { ... }

    ... 
}
```

That reads for me a bit nicer compared to the current `inState<S>(additionalCondition={...})`

but obviously has some disadvantages like:
- exhausting `when` functionality alike (could probably be solved with Lint checks or so)
- DSL contains more and more logic that cannot be unit tested anymore individually and reminds me a bit of too much pollated kotest DSL
- ...

Any thoughts?

gabrielittner:
Closes #310, implementation based on what we discussed in #480

The `condition` block is effectively just a second subclass of `BaseBuilderBlock` for which `InStateBuilderBlock` will provide the entry point. As part of this `inStateWithCondition` and `isInState` overloads with `additionalIsInState` parameter have been deprecated.


sample/shared_code/src/commonMain/kotlin/com/freeletics/flowredux/sample/shared/PaginationStateMachine.kt
```diff
@@ -34,21 +34,15 @@ class InternalPaginationStateMachine(
...
+                condition({ it.canLoadNextPage && it.nextPageLoadingState == NextPageLoadingState.LOADING }) {
+                    onEnter { loadNextPage(it) }
+                }
 
-            inState<ShowContentPaginationState>(additionalIsInState = {
-                it.nextPageLoadingState == NextPageLoadingState.ERROR
-            }) {
-                onEnter { showPaginationErrorFor3SecsThenReset(it) }
-            }
+                condition({ it.nextPageLoadingState == NextPageLoadingState.ERROR }) {
```
sockeqwe, line 42: Reads way better now 👏

sockeqwe (approved):

sockeqwe:
Should we also update the docummentation while doing the code changes or should we do it as a last task when also  `withIdentity` is implemented?

sockeqwe:
I can take care of the documentation. Will create another branch / living PR where I will update the docs with all the changes we do now....

gabrielittner:
I wanted to do it once all changes are there, thanks for taking care of it

PR #518, opened 2023-06-19
implement untilIdentityChanges block

### LINKED ISSUE #480: DSL: Add a way to work with"Identity" of states

gabrielittner:
Closes #480 

With all the ground work that already happened in the refactorings this is now pretty simple. `sideEffectIsInState` where we create the `SideEffect.IsInState` based on `SideEffectBuilder.IsInState` now gets the `initialState` as parameter. The new `IdentityBuilderBlock` overrides that function and in addition to the regular isInState check will check that the identity of the given state matches the identity of the initial state. And that is pretty much it, the only other change was to add `untilIdentityChanges` to `InStateBuilderBlock` and `ConditionBuilderBlock`.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
@@ -9,6 +9,14 @@ public class ConditionBuilderBlock<InputState : S, S : Any, A : Any> internal co
     override val isInState: SideEffectBuilder.IsInState<S>,
 ) : BaseBuilderBlock<InputState, S, A>() {
 
+    /**
+     * Anything inside this block will only run while the [identity] of the current state
+     * remains the same. The `identity` is determined by the given function and uses
+     * [equals] for the comparison.
+     *
+     * When the `identity` changes anything currently running in the block is cancelled.
+     * Afterwards a the blocks are started again for the new `identity`.
+     */
     public fun untilIdentityChanges(
         identity: (InputState) -> Any,
         block: IdentityBuilderBlock<InputState, S, A>.() -> Unit,
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
@@ -7,4 +7,14 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 @FlowReduxDsl
 public class ConditionBuilderBlock<InputState : S, S : Any, A : Any> internal constructor(
     override val isInState: SideEffectBuilder.IsInState<S>,
-) : BaseBuilderBlock<InputState, S, A>()
+) : BaseBuilderBlock<InputState, S, A>() {
+
+    public fun untilIdentityChanges(
+        identity: (InputState) -> Any,
+        block: IdentityBuilderBlock<InputState, S, A>.() -> Unit,
+    ) {
+        sideEffectBuilders += IdentityBuilderBlock<InputState, S, A>(isInState, identity)
+            .apply(block)
+            .sideEffectBuilders
+    }
+}
--- a/flowredux/api/flowredux.api
+++ b/flowredux/api/flowredux.api
@@ -30,6 +30,7 @@ public final class com/freeletics/flowredux/dsl/ChangedStateKt {
 }
 
 public final class com/freeletics/flowredux/dsl/ConditionBuilderBlock : com/freeletics/flowredux/dsl/BaseBuilderBlock {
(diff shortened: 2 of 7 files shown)


flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
```diff
@@ -7,4 +7,14 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 @FlowReduxDsl
 public class ConditionBuilderBlock<InputState : S, S : Any, A : Any> internal constructor(
     override val isInState: SideEffectBuilder.IsInState<S>,
-) : BaseBuilderBlock<InputState, S, A>()
+) : BaseBuilderBlock<InputState, S, A>() {
+
+    public fun untilIdentityChanges(
+        identity: (InputState) -> Any,
```
sockeqwe, line 13: Should we add some type alias (IndentityCheck) with some proper java doc that identity is checked with `==` operator? Typealias just to not have to repeat the javadoc in `InStateBuilderBlock` and in `ConditionBuilderBlock` (but maybe without typealias and just copy & pase javadoc is doing the job too since it is only 2 places). 



I am also wondering if we should define a lambda instead where one could specify how identity works? What I mean by that is a way to compare to states (previous and current state) if they have the same identity: `checkSameIdentity: (InputState, InputState) -> Boolean` With that way one could also use other operators to compare identity, i.e. `===` like `checkSameIdentity = { previous : Inputstate , current : IntputState -> previous.foo === current.foo }`. Mabye it is an overkill. Just thinking out loud ...

sockeqwe (approved):

flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
sockeqwe: The more I think about it, the more I think it is overkill to have `checkSameIdentity: (InputState, InputState) -> Boolean` 😄 .


Probably  this would be another operator something like `untilChanged` or whatever. But I think, until there is no usecase, we can keep your proposal

PR #519, opened 2023-06-19
introduce collectWhileInStateBasedOnState

### LINKED ISSUE #186: Overload for collectWhileInState with Builder
Currently we support `collectWhileInState` with `flowBuilder` as shown bellow:

```kotlin
 fun <T> collectWhileInState(
        flowBuilder: (Flow<InputState>) -> Flow<T>,  
        handler: InStateObserverHandler<T, InputState, S>
 )
 ```

I would like to add  an overload  with 

```kotlin
 fun <T> collectWhileInState(
        flowBuilder: (InputState) -> Flow<T>, // just InputState , not Flow<InputState> as above
        handler: InStateObserverHandler<T, InputState, S>
 )
```

because I personally have more need for this variation of flowBuilder ( `(InputState) -> Flow<T>`)  compare to the current one `(Flow<InputState>) -> Flow<T>`. 

Obviously, that causes a clash on the JVM  because both methods have same signature.

Thus we would need to rename one.

I would propose the following:

- My assumption is that most people have use case for the version with `(InputState) -> Flow<T>`  . Thus I would keep that one under the `collectWhileInState` name.
```kotlin
 fun <T> collectWhileInState(
        flowBuilder: (InputState) -> Flow<T>, // just InputState , not Flow<InputState> as above
        handler: InStateObserverHandler<T, InputState, S>
 )
```

- I would rename the one with `Flow<InputState>) -> Flow<T>` to `collectWhileInStateWithGenericBuilder()`:
 ```kotlin
 fun <T> collectWhileInStateWithGenericBuilder(
        flowBuilder: (Flow<InputState>) -> Flow<T>,  
        handler: InStateObserverHandler<T, InputState, S>
 )
```

Any thoughts?

cc @gabrielittner

### PR DESCRIPTION
**Pull request by gabrielittner**
Closes #186 

Introduces the simple flow builder to `collectWhileInState`. This together with `untilIdentityChanges` makes `collectWhileInState` the `(Flow<S>) -> Flow<T>` obsolete and it was deprecated.

Unfortunately the new method has the same signature as the now deprecated one, so I called it `collectWhileInStateBasedOnState`. Not super happy with that name, so suggestions for alternatives are welcome. In an eventual 2.0 we can remove the deprecated method and then rename this one to `collectWhileInState`.

### CODE REVIEW

File: flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/BaseBuilderBlock.kt [hunk 1]
```diff
@@ -151,6 +151,31 @@ public abstract class BaseBuilderBlock<InputState : S, S : Any, A : Any> interna
...
+     * is applied.
+     */
+    public fun <T> collectWhileInStateBasedOnState(
+        flowBuilder: (InputState) -> Flow<T>,
+        executionPolicy: ExecutionPolicy = ExecutionPolicy.ORDERED,
+        handler: suspend (item: T, state: State<InputState>) -> ChangedState<S>,
+    ) {
+        sideEffectBuilders += SideEffectBuilder(isInState) { initialState ->
+            CollectWhile(
+                isInState = sideEffectIsInState(),
```
gabrielittner, line 171: Need to update this when the identity pr is merged

```suggestion
                isInState = sideEffectIsInState(initialState),
```

sockeqwe:
> Unfortunately the new method has the same signature as the now deprecated one, so I called it collectWhileInStateBasedOnState. Not super happy with that name, so suggestions for alternatives are welcome. In an eventual 2.0 we can remove the deprecated method and then rename this one to collectWhileInState.

Since it is quite an advanced operator, I would be very surprised if outside of Freeletics a lot of people would complain if we would break backwards compatibility. I am also not super happy with the name, therefore I am wondering if we cannot simply go with a breaking change by removing the method `collectWhileInState with Flow<S> -> Flow<T>` completely instead of deprecating it and use the same method name for this new collecctWhileInstate with InputState -> Flow<T>`? Are there a lot of breaking changes at Freeletics? Maybe Stadium?

sockeqwe (approved):

gabrielittner:
Staedium is gone and we don't you it in the main app. Went ahead and removed the old method.

PR #520, opened 2023-06-19
add multiplatform support to compose artifact

### LINKED ISSUE #395: Make Compose artifact Compose Multiplatform
At the moment the compose artifact is not targeting compose multiplatform and therefore not usable in compose desktop or compose web. 

A nice enhancement would be to support compose multiplatform for the `:compose` artifact of FlowRedux. Just copy & pasting the `rememberStateAndDispatch()` function definition into compose desktop/web applications is what I do right now (and it works without any modification). Would be nice to have this supported by FlowRedux directly.

gabrielittner:
Closes #395.

This basically switches the project to mpp, replaces the compose runtime with the multiplatform one and adds the compose Gradle plugin. The sample app which remained unchanged shows that everything is still working. All of our existing targets that are also supported by Compose Multiplatform are added.

Had to downgrade Kotlin and because of that also AndroidX Compose because Compose Multiplatform doesn't support 1.8.21 yet.


sockeqwe (approved):
Nice!

PR #540, opened 2023-07-22
update CHANGELOG for 1.2.0

gabrielittner:
I'm planning to cut a release next week, this already adds the changelog for everything that is currently on main.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,61 @@
 Change Log
 ==========
 
-## 1.1.1 **UNRELEASED**
+## 1.2.0 **UNRELEASED**
 
--
+#### DSL additions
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,61 @@
...
+        }
+      }
+    }
+  }
+  ```
+
+#### Internal re-write
+
+- The internals of FlowRedux have been completely rewritten and simplified. The library behavior
+  is now much more consistent and predictable.
```
gabrielittner, line 45: btw, we had 0 test flakes in here since the re-write

sockeqwe: Nice 🎉

PR #547, opened 2023-07-27
allow identity to be null

### LINKED ISSUE #546: Support null values for identity?
Right now `untilIdentityChanged( identity : (InputState) -> Any)` doesn't allow null values as return type, hence support null as  identity. 

But I am wondering if `null` is actually a valid identity.

Use case: maybe something like a "master-detail" UI, i.e. a gmail tablet ui (email client).

```kotlin
data class InboxState(
    val emails : List<Email>
    val selectedEmail : SelectedEmail?
)

data class SelectedEmail(
    val emailId : Int, 
    val details : EmailDetails?
)
```

```
spec {
    inState<InboxState> {
        onEnter { loadEmails() }
        on<EmailSelected> { action, state -> state.mutate {
            copy(selectedEmail = SelectedEmail(
                    emailId = action.selectedEmail.id,
                    details = null
                ))
            } 
        }

        untilIdentityChanged({ state -> state.selectedEmail?.emailId }) { // selectedEmail can be null
            // this block will be canceled and restarted whenever emailId changes
            onEnter { state -> 
                val s = state.snapshot
                if (s.selectedEmail != null) {
                    val details = loadEmailDetails(s.selectedEmail.emailId)
                    state.mutate { 
                        copy(selectedEmail = selectedEmail.copy(details = details))
                    }
                } else {
                    state.noChange()
                }
            }
        }
    }
}
```

What do you think @gabrielittner ?

gabrielittner:
Closes #546 

The tests are basically just a copy of the existing tests with null values.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux/dsl/ConditionBuilderBlock.kt
@@ -18,7 +18,7 @@ public class ConditionBuilderBlock<InputState : S, S : Any, A : Any> internal co
      * Afterwards a the blocks are started again for the new `identity`.
      */
     public fun untilIdentityChanges(
-        identity: (InputState) -> Any,
+        identity: (InputState) -> Any?,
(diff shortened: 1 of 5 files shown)


sockeqwe (approved):
👍

PR #554, opened 2023-08-19
rememberStateAndDispatch: wrap `StateAndDispatch` ctor with remember

hoc081098:
Although `state` and `dispatchAction` are remembered, but StateAndDispatch constructor can be called many times when `rememberStateAndDispatch` are called (because of the outer scope).
This PR wraps `StateAndDispatch` ctor with `remember(state, dispatchAction)` to avoid unnecessary re-creation of `StateAndDispatch`.


gabrielittner (approved):
Thanks 👍

Issue #521, opened 2023-06-19
2.0 todo list


gabrielittner:
There are no plans yet for going to 2.0, but I want to start collecting things that can be cleaned up, so that we won't forget them when 2.0 actually comes.

- remove `JvmName`
- remove deprecated `inState` overloads
- remove deprecated `inStateWithCondition`
- remove deprecated `collectWhileInState`(`Effect`) overloads
- rename `collectWhileInStateBasedOnState` to `collectWhileInState`
- move `FlowReduxStateMachine` to `com.freeletics.flowredux`
- re-visit naming of classes in the dsl package

PR #625, opened 2023-12-18
Use khoshu state-machine as dependency directly

kboyarshinov:
## Description

Instead of intermediate release of `mad` use `khonshu:state-machine` directly.


gabrielittner:
Isn't this a binary incompatible change since we're removing a part of the type hierarchy? That's why I kept it as until now

kboyarshinov:
@gabrielittner yes, you're right. We can bundle it later with next major release. Closing it for now.

gabrielittner:
I've added it to the todo list here https://github.com/freeletics/FlowRedux/issues/521

PR #833, opened 2025-06-13
split State into State and ChangeableState

gabrielittner:
Rename the existing `State` to `ChangeableState` and extract a new `State` base class that only provides access to the `snapshot`. In the next step this base class will be used to replace the standalone `snapshot` parameters in the DSL. Having this will have 2 advantages:
1) A more uniform API as you always get one of our state wrappers.
2) make it more obvious that you are receiving a snapshot. Currently the standalone `snapshot` parameter can be called anything by the user, so if someone calls it `state` or just uses `it` it can lead to wrong unknowingly using it wrong. With this you always do `_.snapshot` making it obvious.


zaream (approved):

kboyarshinov (approved):
LGTM

PR #836, opened 2025-06-13
change State wrapper to be receiver of all DSL handlers

gabrielittner:
All DSL handlers now have either `State<S>` or `ChangeableState<S>` as receiver to create a more uniform API and make the DSL less noisy.

Before:
```kotlin
onEnter { it.mutate { ... } }
onEnterEffect { ... }
on<Action> { action, state -> state.mutate { ... } }
onActionEffect<Action> { action, state -> ... }
// collectWhileInState etc are like the action handlers
```

After:
```kotlin
onEnter { mutate { ... } }
onEnterEffect { ... }
on<Action> { mutate { ... } }
onActionEffect<Action> { ... }
// collectWhileInState etc are like the action handlers
```

- `State` access now always happens through `this`
- Other parameters like an action are always the only parameter and can be easily accessed through `it`.
- Unused parameters can be easily ignored, previously you'd very often have `action, _ ->` for unused state snapshots, `_, state ->` for actions that don't have values or `_, _ ->` for the combination of both creating a lot of noise. This change in particular is very visible in this PR because the tests had this a lot.

### DIFF
--- a/flowredux/src/commonMain/kotlin/com/freeletics/flowredux2/BaseBuilder.kt
+++ b/flowredux/src/commonMain/kotlin/com/freeletics/flowredux2/BaseBuilder.kt
@@ -34,7 +34,7 @@ public abstract class BaseBuilder<InputState : S, S : Any, A : Any> internal con
      */
     public inline fun <reified SubAction : A> on(
         executionPolicy: ExecutionPolicy = ExecutionPolicy.CANCEL_PREVIOUS,
-        noinline handler: suspend (action: SubAction, state: ChangeableState<InputState>) -> ChangedState<S>,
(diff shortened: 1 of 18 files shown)


kboyarshinov (approved):
Nice :shipit:

xiaozhikang0916:
Sorry to comment in a closed thread.

Will this change be released in near future?

gabrielittner:
Hi, yes this will be part of 2.0.0, there will at least an alpha release in the next 2 weeks (were alpha means that there might be further changes before the final 2.0.0). Not sure yet when the final 2.0.0 will be out but I think latest in July

PR #837, opened 2025-06-17
add DSL marker to State in mutators

gabrielittner:
This will prevent 2 things

Calling a mutator inside a mutator:
```kotlin
on<Foo> {
   override {
        mutate { // fails
        }
   }
}
```

Accessing `snapshot` inside a mutator:
```kotlin
on<Foo> {
   override {
        snapshot.copy(...) // fails
   }
}
```
Especially the latter is helpful with avoiding common mistakes were a potentially outdated state is used to construct the new state instead of using the state that is passes as `this` in the mutator. It doesn't solve it completely since you could still build a new state from the snapshot outside of the mutator and then use it but better than what we had before.

Also makes the DSL marker annotation internal.


kboyarshinov (approved):
:shipit:

PR #840, opened 2025-06-20
implement ExecutionPolicy.Throttled(Duration)

gabrielittner:
Adds a new `ExecutionPolicy.Throttled(Duration)`. This allows easily throttling actions or flow emissions in state machines if sources can be too fast. The implementation behaves like RxJava's `throttleFirst` ([marbles](https://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Observable.html#throttleFirst-long-java.util.concurrent.TimeUnit-)) and not like a `debounce`. In practice this means that the first action per time window is immediately handled.

As part of this `ExecutionPolicy` was changed from an enum to an interface because `Throttled` needs to hold a user supplied `Duration`.


kboyarshinov (approved):
👍

PR #855, opened 2025-07-11
skip state emission for effect handlers

### LINKED ISSUE #850: Effect handlers cause state emisssions
Currently when using onEnterEffect or one of the other effect handlers the previous state is emitted again. This is because they just use `noChange()` internally.

gabrielittner:
Closes #850


kboyarshinov (approved):
:shipit:

Issue #548, opened 2023-07-27
Improve hierarchical state machines


sockeqwe:
Something to consider for FlowRedux 2.0 as working with `onEnterStartStateMachine` or `onActionStartStateMachine` is not so convenient, i.e. you dont need to deal with the initial state problem  of the child state machines.

Maybe we could do something like we do with `untilIdentityChanged( { ... } )`.

I am not sure how feasible it is to implement and how much sense it makes, I am just sketching out an idea here:

```kotlin
sealed interface State
object Loading : State
data class ShowContent(items : List<Item>, val otherState : OtherState) : State

selead interface OtherState
object A : OtherState
object B : OtherState
```

then we could have our regular FlowReduxStateMachine but instead of having another FlowReduxStateMachine that manages `OtherState` we could let that be handled by the "parent" FlowReduxStateMachine (i will add a thought about reusability at the end) :

```kotlin
class MyStateMachine : FlowReduxStateMachine<State, Action>( Loading ){

  init {
    spec {
      inState<Loading> {
         onEnter { state -> 
           delay(1000)
           state.override { ShowContent( emptyList(), A) }
        }
     }

   inState<ShowContent> { 

     // new DSL block
     asChildState( { state : ShowContent -> state.otherState }) {  // specify which property of ShowContent state should be treated as a sub state
          // inside asSubState you can use all the regular DSL blocks as if it was the top level spec block of a regular FlowReduxStateMachine
         inState<A> {
            on<Action1> { ... }
            onEnter { ... }
            collectWhileInState { ... }
            condition { ... }
            untilIdentityChanged { ... }

            asSubState(...) // In theory we could support nesting too
         }
   
         inState<B> {     
            ...
         }

      }
    }
  }

}
```

The new DSL block is `asChildState` (better name needed). Basically you can "teach" a FlowReduxStateMachine how to act on a certain property of a State by treating it as a substate. That should come close to something like `onEnterStartStateMachine` or `onActionStartStateMachine` do today but I am not 100% sure if it is a 1-to-1 replacement or not 😄  need to think about it a bit more in detail ... 

One of the nice things of `onEnterStartStateMachine` or `onActionStartStateMachine` is that you can make "reusable" smaller state machines. This is done by having small FlowReduxStateMachines on its own that can be "composed" into bigger state machines and reused when needed. We should aim to have the same advantage with the propose `asChildState` block.
[...]

PR #858, opened 2025-07-17
introduce FlowReduxStateMachineFactory

### LINKED ISSUE #654: [Question] State Collection On Multiple Screens
Is it save co collect state on multiple screens from the same (injected) instance of state machine (potentially on different threads) and using both `collect` and  `rememberState`?

e.g.

```kotlin
//  Collection at custom class
MainScope().launch {
            stateMachine.state.collect { state ->
                
            }
        }

//  Collection at screen
val state = viewModel.stateMachine.rememberState()
```

gabrielittner:
Introduces `FlowReduxStateMachineFactory` as the class to implement and write the DSL, there is a `launchIn(CoroutineScope)` method that then returns a `FlowReduxStateMachine` which holds an active (but lazily started) `StateFlow`. 

The 2 main advantages are:
- It is easier to observe an instance of the state machine in multiple places.
- The internal input actions channel is unique per `FlowReduxStateMachine` instance. This way we don't need the checks in dispatch whether the state machine is being collected. When you obtain an instance through `launchIn` it's active and you can send actions into it. In case it's not being collected right now they are buffered in the channel.

Closes #654 which was the original motivation and highlights the issue a bit more.

Notes:
- The previous `FlowReduxStateMachine` is kept as `LegacyFlowReduxStateMachine` for easier migration (for the next to points in FlowRedux itself, for consumers and for Khonshu)
- compose integration will be updated in the next PR
- sub state machines will be migrated separately, there is also #548 to consider


kboyarshinov (approved):
👍

PR #860, opened 2025-07-21
rewrite compose integration

gabrielittner:
The new `FlowReduxStateMachineFactory` gets a new `produceStateMachine` extension function that returns an active `FlowReduxStateMachine` containing a compose `State`. Advantages over the old approach are that we don't need the extra `StateAndDispatch` class anymore and that inside the implementation we can use the `initialStateSupplier` to create the initial state for `produceState`. The latter means that we immediately have the initial state in the same composition instead of having to wait for a `Flow` emission. 

There is also a `stateMachine` composable function that allows declaring the state machine directly inside a composable function.

The main module is temporarily only supporting compose compatible KMP targets, I'll file a bug on youtrack about this, We're only loosing one watchOS  target (will be supported in compose 1.9), wasm/wasi and the Android native targets though so it's not much of an issue.


kboyarshinov (approved):
:shipit:

PR #893, opened 2025-09-16
use FlowReduxStateMachineFactory for sub state machines

gabrielittner:
Migrates the sub state machine builds to use `FlowReduxStateMachineFactory` instead of `StateMachine`. 

As part of this I cleaned up the API a little bit:
- to reduce the amount of overloads the factory is always passed in a lambda
- to align with the v2 DSL the lambda now has `State<InputState>` as receiver which in practice means you use `snapshot` instead of `it` or a custom name to access the current state when the state machine is created
- more generally there is just one overload with and one overload without `actionMapper`
- removed the default value for `stateMapper` (now `handler` to match other DSL methods), this way it's more consistent and we don't have the uncheck cast anymore

On the implementation side things got a bit simpler and we don't need `CoroutineWaiter` anymore. With `FlowReduxStateMachineFactory` the created `FlowReduxStateMachine` is able to receive actions immediately and they will be buffered internally if the state machine hasn't started yet.


flowredux/src/commonTest/kotlin/com/freeletics/flowredux2/sideeffects/SubStateMachinesMapTest.kt
gabrielittner, line 1: I removed this because the tests only worked by having a fake implementation of `StateMachine`. The functionality is still covered by the `OnActionStartStateMachineTest`

kboyarshinov (approved):
:shipit:

PR #900, opened 2025-09-18
add onEnterLoadSmoothly extension

gabrielittner:
Adds our internal helper as an extension. There are a few changes compared to what we have internally:
- updated parameter naming to make it more obvious what is what
- added parameters for the delays to allow customization
- use `TimeSource` instead of `timeProvider: () -> Long = { System.currentTimeMillis() }`


kboyarshinov (approved):
👍

Issue #880, opened 2025-08-23
Issue and questions about subStateMachine


xiaozhikang0916:
I have an issue and some questions about the `subStateMachine`.

1. In the `onXXXStartStateMachine` apis, they are requiring a `StateMachine` instance, or a lambda returing such instance in the param. And it is impossible in the new FlowRedux2 api now, because we can only access to a StateMachineFactory instance.

The following are some questions about the design of `subStateMachine`.

1. Is it possible to have a stop sign, either from the host state or the sub state, for the sub state machine, to stop and release the running sub state machine and release resources when met. Currently, it is hard to figure out if a sub state machine started by a one-shot action is correctly stopped.
2. I wonder if the state mapper of sub state machine can be `suspend`, or even map the sub state to an action for the host state machine. Or is this thought a violation of some principle?

xiaozhikang0916:
> They are running until either another action of the same kind is received

It should be `another action of the same value` rather than `same kind`, right?

Take your `GithubRepository` Sample as an example. If we add some log to the `MarkAsFavoriteStateMachine` we can find that these state machines are kept alive until the same repo is triggered favourited or unfavourited.

```kotlin
class MarkAsFavoriteStateMachine(
    private val githubApi: GithubApi,
    private val repository: GithubRepository,
): LegacyFlowReduxStateMachine<GithubRepository, Action> {
    override val state: Flow<GithubRepository>
        get() = super.state.onStart { println("$repository Start") }.onCompletion { println("$repository Complete") }
}
```

```
GithubRepository(id=3, name=Repository 3, stargazersCount=30, favoriteStatus=NOT_FAVORITE) Start
GithubRepository(id=4, name=Repository 4, stargazersCount=40, favoriteStatus=NOT_FAVORITE) Start
GithubRepository(id=5, name=Repository 5, stargazersCount=50, favoriteStatus=NOT_FAVORITE) Start
GithubRepository(id=4, name=Repository 4, stargazersCount=41, favoriteStatus=FAVORITE) Start
GithubRepository(id=4, name=Repository 4, stargazersCount=40, favoriteStatus=NOT_FAVORITE) Complete
```

And in our real project, I have a content page and would like to separate the initial load to a sub state machine. It simultaneously loads from database and network asynchronously, emits any of them once value returned, and all works are done after loaded from network.
It is achieveable to close the sub state machine gracefully by designing the host state class carefully, but it would be more convenient with a stop sign in the sub state.

[...]

PR #923, opened 2025-11-03
implement cancelOnState for sub state machines

gabrielittner:
Currently a sub state machine will mostly run endlessly. The only 2 cancel mechanisms are:
- the state machine leaving the containing block (`inState`, `condition` etc)
- for `onActionStart...`: the same action being received another time which will cancel the existing state machine and start a new one

This adds an optional `cancelOnState` lambdam which allows cancelling based on the state emitted by the sub state machine, allowing to write small focused state machines with a terminal state (for example performing some network request and then going to success/failed). 

Addresses #880


kboyarshinov (approved):
🎉

